PR #1, opened 2019-07-30
submition

tsolman:


turboMaCk:
hello @tsolman is this PR ready for review?

turboMaCk:
also don't be afraid to ask any questions if you have some. We're here to help.

Janiczek:
I encourage you to setup your code editor to run `elm-format` on the Elm files after every save - you'll love it :)

Janiczek:
BTW it should probably be a SPA (`Browser.application`) instead of three separate Elm apps

tsolman:
Hello and thank you for your comments :)

tsolman:
Thanks for the review ! It was my first time with elm so still got a lot to learn.

Janiczek:
@tsolman Let us know when it's ready for review :)

breeds.elm
```diff
@@ -0,0 +1,147 @@
...
+breedItemDecoder =
+  JD.map3 Breed
+    (field "id" string)
+    (field "name" string)
+    (field "description" string)
+
+
+breedListDecoder: Decoder (List Breed)
+breedListDecoder =
+  JD.list breedItemDecoder
```
turboMaCk: I would recommend using `elm-format`. it can be feel a bit to opinionated at first but enforcing same formatting across the team is quite beneficial.

index.html
```diff
@@ -0,0 +1,10248 @@
...
+		case 4:
+			// gather nested taggers
+			var xTaggers = x.j;
+			var yTaggers = y.j;
+			var nesting = false;
+
+			var xSubNode = x.k;
+			while (xSubNode.$ === 4)
+			{
+				nesting = true;
```
turboMaCk: this should be in `gitignore` as this is a output of compiler.

breeds.elm
tsolman: i agree as well. A common formatting tool should be inforced.

index.html
tsolman: Sure

PR #2, opened 2019-08-01
First commit

naimis20:
Naim Gaberlo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,32 @@
-# GlobalWebIndex Engineering Challenge
-## Exercise: CatLover
-Create an elm application for cat lovers which is going to build upon thecatapi.com and will have 3 views ... 
-The first view displays a list of 10 random cat images and a button to load more. Clicking on any of those images opens a modal view with the image and the information about the cat’s breed if available. This would be a link to the second view below - the breed detail. The modal should also contain a form to mark the image as your favourite (a part of the third view as well). Make sure you can copy-paste the url of the modal and send it to your friends - they should see the same image as you can see. 
+# GWI Elm
 
-The second view displays a list of cat breeds. Each breed opens a modal again with a list of cat images of that breed. Each of those images must be a link to the image detail from the previous point. 
+We have 3 Pages : Page 1, Page 2 and Page 3 according to the challenge description.
+The Decoders and Contracts modules are used to create Decoders for Json responses as well as Model types 
+Ports is used for local storage to persist favorite cats data
 
-The third view allows you do the following things after providing the necessary information:
-* Display your favourite cats 
-* Mark a specific image as your favourite 
+In order to build the app you have to run the following
+`elm make src/Main.elm --output elm.js`
 
-You can find the API documentation here: https://docs.thecatapi.com/ 
-We give you a lot of freedom in technologies and ways of doing things. We only insist on you using the Elm language. Note that we have omitted a lot of details in this description which we hope you will fill in and thus prove to us that you are aware of industry best practices and that you also follow them. Get creative as much as you want, we WILL appreciate it. You will not be evaluated based on how well you follow these instructions, but based on how sensible your solution will be. In case you are not able to implement something you would normally implement for time reasons, make it clear with a comment. 
+In order to constantly updating current view you have to run the following
+`elm-live src/Main.elm --open -- --output=elm.js`
 
-## Submission
+To save the favorites, a state module is responsible for saving in a Dict structure
 
-Just a make a PR to the current repo!
-Good luck, potential colleague! 
+Used bootstrap (but non overdid it)
 
+Modal was implemented using Tesk9 package
+
+To run the example you have to run on top folder
+`elm reactor`
+and go to http://localhost:8000/index.html to see the application live
+
+### ---Disclaimer--- 
+1. Normally I would split even further the code to Views,Updates,Models,Messages but due to time constraint I was not able to achieve it. 
+2. Internationalization missing
+3. On saving as favorite the following would happen normally:
+   - Initialize the app with actual favorite data from server
+   - Send a post request to server
+4. On second page I believe it is for the better to actually reopen the familiar first page instead of opening a modal on top of the modal
+   - Thus the breedId in Page1 to filter out only cats by breed
+5. Tried not to overengineer but also tried to create a good application architecture infrastructure
+6. Pagination missing
--- a/elm.js
+++ b/elm.js
@@ -1,0 +1,8662 @@
+(function(scope){
+'use strict';
+
+function F(arity, fun, wrapper) {
+  wrapper.a = arity;
+  wrapper.f = fun;
+  return wrapper;
+}
+
+function F2(fun) {
+  return F(2, fun, function(a) { return function(b) { return fun(a,b); }; })
+}
+function F3(fun) {
+  return F(3, fun, function(a) {
+    return function(b) { return function(c) { return fun(a, b, c); }; };
+  });
+}
+function F4(fun) {
+  return F(4, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return fun(a, b, c, d); }; }; };
+  });
+}
+function F5(fun) {
+  return F(5, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return function(e) { return fun(a, b, c, d, e); }; }; }; };
+  });
+}
+function F6(fun) {
+  return F(6, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return function(e) { return function(f) {
+    return fun(a, b, c, d, e, f); }; }; }; }; };
+  });
+}
+function F7(fun) {
+  return F(7, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return function(e) { return function(f) {
+    return function(g) { return fun(a, b, c, d, e, f, g); }; }; }; }; }; };
+  });
+}
+function F8(fun) {
+  return F(8, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return function(e) { return function(f) {
+    return function(g) { return function(h) {
+    return fun(a, b, c, d, e, f, g, h); }; }; }; }; }; }; };
+  });
+}
+function F9(fun) {
+  return F(9, fun, function(a) { return function(b) { return function(c) {
+    return function(d) { return function(e) { return function(f) {
+    return function(g) { return function(h) { return function(i) {
+    return fun(a, b, c, d, e, f, g, h, i); }; }; }; }; }; }; }; };
+  });
+}
+
+function A2(fun, a, b) {
+  return fun.a === 2 ? fun.f(a, b) : fun(a)(b);
+}
+function A3(fun, a, b, c) {
+  return fun.a === 3 ? fun.f(a, b, c) : fun(a)(b)(c);
+}
+function A4(fun, a, b, c, d) {
+  return fun.a === 4 ? fun.f(a, b, c, d) : fun(a)(b)(c)(d);
+}
+function A5(fun, a, b, c, d, e) {
+  return fun.a === 5 ? fun.f(a, b, c, d, e) : fun(a)(b)(c)(d)(e);
+}
+function A6(fun, a, b, c, d, e, f) {
+  return fun.a === 6 ? fun.f(a, b, c, d, e, f) : fun(a)(b)(c)(d)(e)(f);
+}
+function A7(fun, a, b, c, d, e, f, g) {
+  return fun.a === 7 ? fun.f(a, b, c, d, e, f, g) : fun(a)(b)(c)(d)(e)(f)(g);
+}
+function A8(fun, a, b, c, d, e, f, g, h) {
+  return fun.a === 8 ? fun.f(a, b, c, d, e, f, g, h) : fun(a)(b)(c)(d)(e)(f)(g)(h);
+}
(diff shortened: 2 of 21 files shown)


src/Decoders.elm
```diff
@@ -0,0 +1,40 @@
...
+
+
+catListDecoder : Decoder (List Cat)
+catListDecoder =
+    JD.list catDecoder
+
+
+breedDecoder : Decoder Breed
+breedDecoder =
+    JD.map4 Breed
```
turboMaCk: In case you're curious. if you want to decode all the attributes you have commented in the record from JSON you can use applicative style to do that. All that's required is handy function defined as `andMap = JD.map2 (|>` but you can use a package for it:

- resource that explains how this works https://toast.al/posts/code/2016-08-12-elm-applicatives-and-json-decoders.html
- popular package https://package.elm-lang.org/packages/NoRedInk/elm-json-decode-pipeline/latest/

But I think it shouldn't be required for this challange.

turboMaCk:
@naimis20 pls let me know when this will be ready for review. Or if it is say so so I know I can make full review of the code. Thanks :hand:

naimis20:
> @naimis20 pls let me know when this will be ready for review. Or if it is say so so I know I can make full review of the code. Thanks ✋

@turboMaCk yes it is ready for review. Thank you for your time and effort :)

turboMaCk:
great thanks. I'm going to look into this ASAP. That means probably during tomorrow. Thanks for your time spent on it. We usually leave a lot of feedback so we balance the time people spent on working on this by investing our time into evaluating. Don't be affraid if I leave a lot of comments tomorrow, it's not a bad thing.

naimis20:
Great! thank you once again for your time 👍

src/Decoders.elm
```diff
@@ -0,0 +1,40 @@
+module Decoders exposing (breedDecoder, breedListDecoder, catDecoder, catListDecoder, weightDecoder)
+
+import Contracts exposing (..)
+import Json.Decode as JD exposing (Decoder, field, int, string)
+
+
+catDecoder : Decoder Cat
+catDecoder =
```
Janiczek: Seems the consensus in the Elm community is to group types and functions like this:
```
- Cat (module)
  - Cat (type)
  - decoder
  - other functions operating on Cats
- Breed (module)
  - Breed (type)
  - decoder
  - other functions operating on Breeds
```
rather than the way it is now. But it's nothing too important here. Still it would be great to hear your thoughts on this!

src/Page/Page2.elm
```diff
@@ -0,0 +1,144 @@
+module Page.Page2 exposing (Model, Msg(..), Status(..), getBreeds, init, subscriptions, toState, update, view)
```
Janiczek: There is probably a better, more descriptive name for these `PageX` modules lurking somewhere :upside_down_face:

src/Decoders.elm
naimis20: I see your point and I believe it should be more clear i.e CatModule to resolve potential problems due to naming issues

src/Page/Page2.elm
turboMaCk: true but requirements also call them pageN so this is inhereted from there.

naimis20: I just splitted the app in 3 pages and this is the reason for such names. In a normal scalable application I would not give such names of course. :)

src/State.elm
```diff
@@ -0,0 +1,63 @@
+module State exposing (State(..), addMostFavorite, addToFavorites, getFavorites, getMostFavorite, navKey, removeFromFavorites)
```
Janiczek: Can you explain the reasoning behind having this `State` type in addition to the `Main.Model`?
eg. why not factor it out if it's in all the possible states and pages:
```elm
-- Main.elm
type alias Model =
  { key : Nav.Key
  , favorites : Dict String Cat
  , mostFavorite : Cat
  , page : Page
  }

type Page
  = Redirect
  | Page1 Page1.Model
  | Page2 Page2.Model
  | Page3 Page3.Model

-- PageX.elm
type alias Model =
  -- doesn't contain State, as the top-level Model contains it
  { status : Status
  , ...
  }
```

naimis20: The state is considered a state mechanism for the whole application whereas the Model is considered the individual page state mechanism

Janiczek:
I think there might be a small misunderstanding of requirements (list of favorites vs one most favorite - I believe the reqs wanted just the list of favorites?), but otherwise the code is pretty fine. You show an understanding of Elm (custom types, SPA, ...) and you've even taught me that tesk9/modal package exists! :)

If this was more long-term code, I'd probably be bitching a lot more about some aspects of the PR (BreedId can be opaque etc.) but that's not the purpose of the challenge. Nice work!

naimis20:
> I think there might be a small misunderstanding of requirements (list of favorites vs one most favorite - I believe the reqs wanted just the list of favorites?), but otherwise the code is pretty fine. You show an understanding of Elm (custom types, SPA, ...) and you've even taught me that tesk9/modal package exists! :)
> 
> If this was more long-term code, I'd probably be bitching a lot more about some aspects of the PR (BreedId can be opaque etc.) but that's not the purpose of the challenge. Nice work!

Thank you mate :) Yes I have understood that both the favorites AND the most favorite would be needed, thus the code above. 
As I wrote in the disclaimer I tried to keep it as simple as possible but also provide some kind of application architecture. In case this was a long term project I believe that over 75% of the current code would be different to support large application scalability.

src/Main.elm
```diff
@@ -0,0 +1,200 @@
...
+
+        Just Route.Page2 ->
+            Page2.init state maybeRoute
+                |> updateWith Page2 GotPage2Msg model
+
+        Just Route.Page3 ->
+            Page3.init state maybeRoute
+                |> updateWith Page3 GotPage3Msg model
+
+        _ ->
```
janjelinek: Currently, this is `Just Root ->` and I would rather don' use the wildcard `_ ->` in cases like this. It guarantees you help from the compiler when something is changed in future.

naimis20: I completely understand.

src/Page/Page1.elm
```diff
@@ -0,0 +1,206 @@
...
+                    ( { model | status = Failure }, Cmd.none )
+
+        ModalCreate modalMsg ->
+            let
+                ( modalModel, cmdMsg ) =
+                    ModalView.update modalMsg model.modal model.selectedItem
+            in
+            ( { model | status = ModalStatus, modal = modalModel }
+            , Cmd.batch
+                [ getCats model
```
janjelinek: Why? It causes reload of all already shown cats under opened modal.

src/State.elm
janjelinek: I don't understand beneffits of this module. Probably it's caused by missleading `Favorites` name in type definition. This is not only for favorites, right? Why you don't use record?
```
type alias State = { favorites : (Dict String Cat) , mostFavoriteCat : Cat, navigationKey : Nav.Key }
```

styles/styles.css
```diff
@@ -0,0 +1,3 @@
+.nav-item.active.list-group-item .nav-link{
+  color:white !important;
```
janjelinek: 🙈

src/Route.elm
```diff
@@ -0,0 +1,64 @@
...
+                Just (Page1 breedId) ->
+                    [ "page1", BreedId.toString breedId ]
+
+                Just Page2 ->
+                    [ "page2" ]
+
+                Just Page3 ->
+                    [ "page3" ]
+
+                _ ->
```
janjelinek: `s/_/Nothing/`

src/Page/Page2.elm
```diff
@@ -0,0 +1,144 @@
...
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    case msg of
+        GetRequest ->
+            ( { model | status = Loading }, getBreeds )
+
+        GetResponse result ->
+            case result of
+                Ok url ->
```
janjelinek: `url`? Really?

README.md
```diff
@@ -1,19 +1,32 @@
...
+Used bootstrap (but non overdid it)
 
+Modal was implemented using Tesk9 package
+
+To run the example you have to run on top folder
+`elm reactor`
+and go to http://localhost:8000/index.html to see the application live
+
+### ---Disclaimer--- 
+1. Normally I would split even further the code to Views,Updates,Models,Messages but due to time constraint I was not able to achieve it.
```
janjelinek: I personally don't like it anyway. I don't understand why so many examples and packages are done this way, because for me is much harder to navigate between files than in a single file. But maybe it's just because I'm used to having it in a single file rather than spread everywhere in the folder 🤔

src/Main.elm
```diff
@@ -0,0 +1,200 @@
...
+    | Page1 Page1.Model
+    | Page2 Page2.Model
+    | Page3 Page3.Model
+
+
+init : () -> Url -> Nav.Key -> ( Model, Cmd Msg )
+init _ url navKey =
+    let
+        initCat =
+            { url = "https://cdn2.thecatapi.com/images/5eg.jpg"
```
janjelinek: Not sure why you choose to have `Cat` instead of `Maybe Cat` for this. It's kinda weird to have the most favourite cat for the first run, but I consider this as a small easter egg from you, so 🤷‍♂ :)

src/Route.elm
```diff
@@ -0,0 +1,64 @@
...
+import Url exposing (Url)
+import Url.Parser exposing ((</>), Parser, int, map, oneOf, s, string)
+
+
+
+type Route
+    = Page1 BreedId
+    | Page2
+    | Page3
+    | Root
```
janjelinek: Seems like `Root` is basically `Page1 "-"` for you in some places. Maybe you shod rethink `Route` type or it's usage.

src/State.elm
```diff
@@ -0,0 +1,63 @@
...
+import Contracts exposing (..)
+import Dict exposing (Dict)
+
+
+type State
+    = Favorites Nav.Key (Dict String Cat) Cat
+
+
+navKey : State -> Nav.Key
+navKey state =
```
janjelinek: You are mixing naming conventions. Below you have `get...` but not for this one.

src/Page/Page1.elm
naimis20: I actually considered this a feature.

src/State.elm
naimis20: I wrote it in this manner so that it could be expandable for possible future state needed features

README.md
naimis20: So the way you are developing is module per feature/component? How do you usually split a large app?

src/Main.elm
naimis20: Actually I really liked the pic and this is the reason I did not use a Maybe Cat :)

src/Route.elm
naimis20: Correct, I started with blank but later due to challenge purposes the blank page seemed like an overkill and I should have removed it from everywhere.

src/Main.elm
janjelinek: That's reasonable enough for me 👍 😹

README.md
Janiczek: yeah, every page/component module has its own Model, Msg, view, update, subscriptions etc. (if needed)

turboMaCk: agree

elm.js
```diff
@@ -0,0 +1,8662 @@
...
+			var subPatches = [];
+			_VirtualDom_diffHelp(x.k, y.k, subPatches, 0);
+			subPatches.length > 0 && _VirtualDom_pushPatch(patches, 1, index, subPatches);
+			return;
+
+		case 4:
+			// gather nested taggers
+			var xTaggers = x.j;
+			var yTaggers = y.j;
+			var nesting = fal
```
turboMaCk: this file should be gitignored as it is compiler output.

elm.json
```diff
@@ -0,0 +1,29 @@
...
+    "elm-version": "0.19.0",
+    "dependencies": {
+        "direct": {
+            "elm/browser": "1.0.1",
+            "elm/core": "1.0.2",
+            "elm/html": "1.0.0",
+            "elm/http": "2.0.0",
+            "elm/json": "1.1.3",
+            "elm/url": "1.0.0",
+            "tesk9/modal": "5.0.1"
```
turboMaCk: I'm not familiar with this dependecy, what is it good for? (I've checked it but I'm interested in hearing why you used it and what you think about it).

src/Page/Page2.elm
turboMaCk: it might be url of the picture?

styles/styles.css
turboMaCk: :see_no_evil:

src/Page/Page1.elm
turboMaCk: I agree with @janjelinek those constant reloads looks more like glitches.

elm.json
naimis20: You mean about tesk9 right? It is an actual implementation of the standard of the modal, however it does not expose certain messages that would make your life easier (i.e ModalClose msg)

src/Page/Page2.elm
naimis20: I have not seen this comment - no this is just wrong naming. It is actually a list of breed objects

src/Page/Page1.elm
naimis20: The glitch is on Modal Open functionality. 
I added this functionality on Modal Close functionality to auto update the cats.

PR #3, opened 2019-08-21
Challenge submission

thsarig:


turboMaCk:
hello @thsarig :wave: 

Is this PR ready for review?

thsarig:
Hello @turboMaCk .
Yes, I will not add something more. 
Thank you for your time

README.md
```diff
@@ -1,19 +1,14 @@
...
-
-Just a make a PR to the current repo!
-Good luck, potential colleague! 
+In order to run the app you have to run the following after installing the create-elm-app
+`elm-app start`
 
+### ---Notes--- 
+1. Due to the lack of time and the steep learning curve I did not implement the third view and the favourite cats functionality. Also the UI is too simple and I use the bootrstrap css  in order to use some css classes for a responsive image list and modal.
+2. Also the second view is a list of the cats of the same breed and not a list of the breeds which will trigger a modal then with the cats list.
+3. I tried to follow the one file approach except of my models. HTTP, routes, decoders are all in the same Main.elm file.
```
turboMaCk: @gnikolaropoulos this contains information about what is implemented and what isn't

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+-- MODEL
+
+
+type alias Model =
+    { cats : List Models.Cat
+    , selectedCat : Maybe Models.Cat
+    , showLoading : Bool
+    , showModal : Bool
+    , key : Nav.Key
+    , route : Maybe Route
```
turboMaCk: I would personally consider having cats as `Dict` where key is cat id and selected cat as `Maybe Id` so that when any updates to cat are hapening there doesn't need to be code to synchronize changes between cat in list and selected state. But even this model has some advantages.

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+    | ToggleModal (Maybe Models.Cat)
+    | ChangedUrl Url
+    | ClickedLink Browser.UrlRequest
+
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    case msg of
+        MorePlease ->
+            ( { model | showLoading = True }, getCatList )
```
turboMaCk: While this is fine usually more usaful way to model loading state is to use [RemoteData](https://package.elm-lang.org/packages/krisajenkins/remotedata/latest/RemoteData)

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+        Nothing ->
+            span [] []
+
+        Just breeds ->
+            div [] (List.map renderABreed breeds)
+
+
+renderABreed : Breed -> Html Msg
+renderABreed breed =
+    a [ href ("/breed/" ++ breed.id), class "d-block" ] [ text breed.name ]
```
turboMaCk: [`Url.Builder`](https://package.elm-lang.org/packages/elm/url/latest/Url-Builder) from url package offers nicer way to build urls like this.

src/Models.elm
```diff
@@ -0,0 +1,34 @@
...
+    , breeds : Maybe (List Breed)
+    , categories : Maybe (List Category)
+    }
+
+
+type RemoteData error value
+    = NotAsked
+    | Loading
+    | Failure error
+    | Success value
```
turboMaCk: I don't see this being used.

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+
+
+
+-- MODEL
+
+
+type alias Model =
+    { cats : List Models.Cat
+    , selectedCat : Maybe Models.Cat
+    , showLoading : Bool
```
jhrcek: Instead of having showLoading Bool it might make more sense to use `RemoteData` which you declared in Models but you're not using it here (using the maxim "making impossible states impossible"). I think you don't have to be afraid to reuse [remotedata](https://package.elm-lang.org/packages/krisajenkins/remotedata/latest/) instead of implementing your own flavor of that data type.

src/Main.elm
jhrcek: Does wrapping the route in Maybe have some disadvantages? For example will the app be well behaved if user enters an invalid URL? Or would it make sense to redirect user e.g. to home page whenever he enter invalid url manually?

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+
+        Loading ->
+            ( { model | showLoading = True }, Cmd.none )
+
+        GotList result ->
+            case result of
+                Ok catlist ->
+                    ( { model | showLoading = False, cats = catlist }, Cmd.none )
+
+                Err _ ->
```
jhrcek: It would be nice to give some feedback in the UI that loading of images was not possible..

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+                        ]
+                    ]
+                ]
+
+
+renderBreeds : Maybe (List Models.Breed) -> Html Msg
+renderBreeds breed =
+    case breed of
+        Nothing ->
+            span [] []
```
jhrcek: Nitpick: rendering `Html.text ""` rather than empty span might be preferable

src/Main.elm
turboMaCk: I guess Nothing serves as 404 route. In platform we have explicit `NotFound` route. I think it makes sense to rather have ability to show error that page don't exist rather than redirecting to different route.

src/Main.elm
```diff
@@ -0,0 +1,316 @@
...
+    ( initialModel url key, getCatList )
+
+
+
+-- UPDATE
+
+
+type Msg
+    = MorePlease
+    | Loading
```
janjelinek: What was the original purpose of this `Msg`? Seems like you forgot to delete it because I can't find where it is fired.

src/main.css
```diff
@@ -0,0 +1,51 @@
...
+.modal {
+  display: inherit;
+}
+
+.modal-body img {
+  width: 90%;
+  max-width: 500px;
+}
+
+#modal.open {
```
janjelinek: This is not the best way how to deal with styles. Mixing together `id` and `class` selectors and even more for different elements 🙈I would recommend you to stop using `id` selectors at all for styling.

src/Main.elm
Janiczek: Context: http://blog.jenkster.com/2016/06/how-elm-slays-a-ui-antipattern.html

thsarig:
First of all I want to thank you all for your constructive comments and your reviews. I will take them into consideration in a next Elm application. I am sorry about the missing start command from the README.md file, I thought it was a common practice about Elm applications (the boilerplate library that I used).
The best thing I saw until now is that magic Elm compiler with its helpful error messages. I had a tough time at the start due to the different format and the build process which I should follow , but after that it was quite OK. Also the elm debugger is very useful and reminds me the Redux Dev Tools I use already in Javascript applications. I liked also the strict approach to functional paradigm which is inside its core process and not through a library like RxJS which I used in Angular appliactions. Even the learning curve which is a bit annoying at the start for everything new , I enjoyed it a lot.

PR #4, opened 2019-09-16
react-redux

axypas:
There is only one commit because initially I thought that the project should be delivered using the link of greenhouse.


README.md
```diff
@@ -1,19 +1,33 @@
...
-* Mark a specific image as your favourite 
+## What is done
+The application implements everything that was mentioned in the requirements.
 
-You can find the API documentation here: https://docs.thecatapi.com/ 
-We give you a lot of freedom in technologies and ways of doing things. We only insist on you using the Elm language. Note that we have omitted a lot of details in this description which we hope you will fill in and thus prove to us that you are aware of industry best practices and that you also follow them. Get creative as much as you want, we WILL appreciate it. You will not be evaluated based on how well you follow these instructions, but based on how sensible your solution will be. In case you are not able to implement something you would normally implement for time reasons, make it clear with a comment. 
+## What is missing
 
-## Submission
+There are no unit tests :(. While I believe that unit tests are very important and part of industry best practices, I have not added any due to time reason.
```
renra: Hello @axypas 

Thanks for your submission. On the topic of tests, how would you approach testing the app? What would you test, what would you rather not test? What makes a good test suite in your opinion?

README.md
```diff
@@ -1,19 +1,33 @@
...
 
-Just a make a PR to the current repo!
-Good luck, potential colleague! 
+The CSS is very basic, most of the CSS relies on Bootsrap.
+
+A strong typing system such as Typescript would be beneficial if the project is meant to grow.
```
gnikolaropoulos: Could you explain you potential choices?

src/modules/search/containers/Cats.js
```diff
@@ -0,0 +1,99 @@
...
+      {isLoading && <div>Loading...</div>}
+      {error && (
+        <div className="error">
+          An error has occured, please refresh the page!
+        </div>
+      )}
+      <CatGallery cats={cats} handleClick={selectCat} />
+      <Button onClick={fetchCats} className="button">
+        More
+      </Button>
```
gnikolaropoulos: How would you handle the case where we want more cats to load on the screen instead of re-rendering this page with new cats?

src/redux/rootReducer.js
```diff
@@ -0,0 +1,10 @@
+import { combineReducers } from "redux";
+import { reducer as searchReducer } from "../modules/search/ducks";
+import { reducer as breedsReducer } from "../modules/breeds/ducks";
+import { reducer as favouritesReducer } from "../modules/favourites/ducks";
+
+export default combineReducers({
+  searchReducer,
+  breedsReducer,
+  favouritesReducer
+});
```
gnikolaropoulos: If we needed to reset the Redux state say to an initial empty state or a default set of cats, how would handle this through reducers?

src/registerServiceWorker.js
```diff
@@ -0,0 +1,117 @@
+// In production, we register a service worker to serve assets from local cache.
```
gnikolaropoulos: I see that this is more or less the default implementation of the create-create-app. Have you ever used serviceworkers in production? How did you deal with major platform updates?

src/App.js
```diff
@@ -0,0 +1,44 @@
...
+          <Route path="/" exact component={Cats} />
+          <Route path="/breeds" exact component={Breeds} />
+          <Route path="/favourites/" component={Favourites} />
+        </div>
+      </Router>
+    );
+  }
+}
+
+export default App;
```
gnikolaropoulos: Overall I like the architecture, I think it's clean. Module names and naming in general seem consistent and follow the standard practices. Liked the fact of using history to keep permanent links to modal. All in all good job, simple and clean, I would have preferred some more advanced css, caching and testing.

Last question => Why choose Redux instead of Mobx?

README.md
axypas: ## Potential choices

To add static typing, one could use TypeScript or Flow.js that let you write JavaScript. An interesting alternative is Reason that let you write statically typed code that is based on OCaml and has a syntax close to JavaScript.

- TypeScript is the oldest and the most used, the ecosystem is huge and a lot of libraries provide types for it. The editor support is great and it is actively maintained.

 - Flow.js used to be more powerful in term of types than TypeScript but TypeScript is now on paar. While Flow.js let you add annotation to JavaScript and as such does not need compilation, TypeScript is a superset of JavaScript that needs to be compiled. The overhead of typescript might be significant depending on the size of the project and how it is configured. The editor support is not too good, except if you use Atom.

- Reason is the latest and more powerful from the three, it is less verbose and  100% sound typing. The compiler is blazing fast and it is backed by Facebook (Flow.js is also backed by Facebook). I don’t have experience yet with Reason, the editor support for Visual Studio Code seems very good.

## Which one to choose?
It depends on the team and needs some experimentation. Among TypeScript and Flow.js, I would personally favor TypeScript over Flow.js since the community is much bigger and Facebook (creator of Flow.js) is pushing towards Reason.

Among Reason and Typescript, developers familiar with Elm (cousin of Reason) would probably be keen to use Reason. There is more risk to choose Reason right now due to the fact that BuckleScript (compiler of Reason to JS) is maintained mostly by one developer, on the other side Facebook has the resources to maintain it. Also the Reason community is looking to compile natively for the mobiles which is interesting.

src/modules/search/containers/Cats.js
axypas: This is a one line change as you can see in https://github.com/axypas/platform2.0-elm-challenge/pull/1/files

src/registerServiceWorker.js
axypas: I have just played with it, I don't have experience with it on production.

src/App.js
axypas: Thank you @gnikolaropoulos for your kind comment.

Why choose Redux instead of Mobx?

The paradigm is different, Redux is functional and Mobx is object-oriented, I guess I favor functional programming :).

In more details:

## Mobx

### Advantages:
- Less verbose
- Performant out of the box
- Design decisions are mostly taken for the developer

### Disadvantage
- Harder to test than pure functions
- Harder to scale
- Quite difficult to debug when the project becomes big due to the reactions
- Breaking changes in the API (happened once, it affected a small portion of our code )
- No support for older browser in the last version
- More concepts to understand when you need advanced functionalities

## Redux

### Advantages:
- Simple to code and to understand what happens
- Easy to debug
- Easy to test
- Easy to refactor
- Excellent documentation

### Disadvantages
- Need to choose some extra libraries (Immutables.JS, action creator, Reselect, normalizing state shape)
- More code needed to be efficient

Overall, Mobx feels like magic when you don't face any issues (and pain when you do) while Redux feels dumb and simple (KISS).

src/redux/rootReducer.js
axypas: You can set the initial state using `preloadedState` in the `combineReducers` or set it in the reducer (ex: `function reducer(state = initialState, action) {`. I have done it in the reducer.

README.md
axypas: Hello @renra!
Thank you for reviewing my pull-request!
 
So the following answers concern the testing of the front-end only.
 
## On the topic of tests, how would you approach testing the app?
I would follow the test pyramid, so a lot of unit tests, some integration tests and few UI tests. To know the test coverage, I would use Jest. The tests would run during the build of the pull-request with a rule preventing merging if tests are failing in the CI.
 
### Unit tests
Concerning the React components, I would have snapshot tests for the presentational components using Jest to make sure that the structure does not change unexpectedly. If there are React components with logic, they could be tested with Enzyme or react-testing-library. 
 
Concerning the reducers, the actions and the middleware, I would test them using Jest. In terms of importance, the middleware might affect all actions, then the reducers have most of the logic, then the actions that are complex. 
 
 
### Integration tests
I would add some contract tests to make sure that the app does not break if the api changes.
 
 
### UI tests
A few UI tests should cover the core functionality of the application. UI tests can also help to prevent regressions in the UI. Some tools such as happo.io or Chromatic are some options. For end-to-end tests, I would go for Cypress that run tests quickly and waits automatically preventing the tests to be flaky.
 
 ## What would you test, what would you rather not test?
Some form of testing is always done, either manually or automatically. Concerning the automatic tests, all code that contains logic should be tested. I would try to have a good unit test coverage and add a unit test for every bug reported (or another form of test if the unit test cannot catch the bug).
 
I would rather not test code that is trivial (getters, very simple actions) or third-party code (test only the behavior of our code that is using it). In term of E2E tests, I would limit them to the core functionality of the application (login, payment, in our case the search page).
 
## What makes a good test suite in your opinion?

- **Fast**: The test suite is bypassed otherwise in the pre-commit
E2E tests that are slow are rarely run, creating a burden.
The tests should be run very often, during the development, on pre-commit, during the build in the CI.
- **Complete**: Test all the possible branches when checking using branch testing.
Will be faster and safer to refactor if tests are complete.
- **Reliable**: Flaky tests create false alarm, reduce the trust in the suite and waste time to make them reliable.
- **Isolated**: Tests should not affect the other tests. The tests should not have to run in a specific sequence to work.
- **Maintainable**: The tests should be as clean as the rest of the code.
The tests should be easy to read.

axypas: @gnikolaropoulos @renra Thank for for the interesting and thoughtful questions. I have jsut finished to answer all of them, I am looking forward hearing from you.

src/modules/breeds/containers/Breeds.js
```diff
@@ -0,0 +1,85 @@
...
+const mapDispatchToProps = (dispatch, ownProps) => {
+  return {
+    fetchBreeds: () => dispatch(fetchBreeds()),
+    fetchBreedsIfNeeded: () => dispatch(fetchBreedsIfNeeded()),
+    selectBreed: breed => dispatch(selectBreed(breed)),
+    unselectBreed: () => {
+      dispatch(selectBreed(undefined));
+    },
+    selectCat: cat => {
+      const { history } = ownProps;
```
turboMaCk: you could have deconstructed history directly in the argument above but I understand it might not be scalable. Why not to use `ownProps.history.push` though as it's used in one place anyway. (I'm more commenting so you know I was really reviewing it though :joy: )

src/modules/breeds/ducks.js
```diff
@@ -0,0 +1,124 @@
...
+        isInitialized: true
+      };
+    }
+    case FETCH_BREEDS_SUCCESS: {
+      const { breeds } = action.payload.variables;
+      return {
+        ...state,
+        isLoading: false,
+        error: undefined,
+        breeds: breeds.map(breed => ({ ...breed, src: breed.url }))
```
turboMaCk: I don't like the fact that you need to manually add `src` attribute on this place as well as the function above (Breed.js:16). Maybe this function should be called in view rather than in here or at least extracted to the common function in case you want to cache it rather than perform this map on every render. Also you can probably avoid using `src` alltogehr and use `url` attribute directly.

src/modules/breeds/ducks.js
```diff
@@ -0,0 +1,124 @@
...
+        selectedBreedId: action.payload.variables.breedId
+      };
+    }
+    case FETCH_CATS_PER_BREED_SUCCESS: {
+      const { cats } = action.payload.variables;
+      return {
+        ...state,
+        isLoading: false,
+        error: undefined,
+        cats: cats.map(cat => ({ ...cat, src: cat.url }))
```
turboMaCk: another like that.

src/modules/favourites/containers/Favourites.js
```diff
@@ -0,0 +1,48 @@
...
+import CatGallery from "components/CatGallery";
+
+function mapStateToProps(state) {
+  const favouriteState = state.favouritesReducer;
+  const { favourites, error, isLoading } = favouriteState;
+  return {
+    error,
+    isLoading,
+    favourites
+  };
```
turboMaCk: why not to define this as `const mapStateToProps => { favoritesReducer } => favoritesReducer`. It doesn't seem to be doing much else, does it?

src/modules/favourites/ducks.js
```diff
@@ -0,0 +1,105 @@
...
+    case FETCH_FAVOURITES_SUCCESS: {
+      const { favourites } = action.payload.variables;
+      return {
+        ...state,
+        isLoading: false,
+        error: undefined,
+        favourites: favourites.map(f => ({
+          ...f.image,
+          favouriteId: f.id,
+          src: f.image.url,
```
turboMaCk: another copying of `url` to `src`

src/modules/search/components/CatModal.js
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import { Button, Image, Modal } from "react-bootstrap";
+
+const concatBreedNames = breeds =>
+  breeds && breeds.length > 0
```
turboMaCk: what does `breeds.length > 0` part of condition do? I believe it's redundant.

src/modules/breeds/ducks.js
```diff
@@ -0,0 +1,124 @@
+const initialSate = {
```
turboMaCk: shy are those modules called `ducks`???

src/modules/search/containers/Cats.js
```diff
@@ -0,0 +1,99 @@
...
+
+import CatGallery from "components/CatGallery";
+import Modal from "../components/CatModal";
+import {
+  fetchCats,
+  fetchCatsIfNeeded,
+  selectCat,
+  selectCatById,
+  markAsFavorite
+} from "../ducks";
```
turboMaCk: fetchCats from ducks :joy:

src/registerServiceWorker.js
turboMaCk: I think this file was produced by whatever was used to lay down the boilerplate

yarn.lock
```diff
@@ -0,0 +1,10654 @@
...
+
+"@types/eslint-visitor-keys@^1.0.0":
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/@types/eslint-visitor-keys/-/eslint-visitor-keys-1.0.0.tgz#1ee30d79544ca84d68d4b3cdb0af4f205663dd2d"
+  integrity sha512-OCutwjDZ4aFS6PB1UZ988C4YgwlBHJd6wCeQqaLdmadZ/7e+w79+hbMUFC1QXDNCmdyoRfAFdm0RypzwR+Qpag==
+
+"@types/istanbul-lib-coverage@*", "@types/istanbul-lib-coverage@^2.0.0":
+  version "2.0.1"
+  resolved "https://registry.yarnpkg.com/@types/istanbul-lib-coverage/-/istanbul-lib-coverage-2.0.1.tgz#42995b446db9a48a11a07ec083499a860e9138ff"
+  integrity sha512-hRJD2ahnnpLgsj6KWMYSrmXkM3rm2Dl1qkx6IOFD5FnuNPXJIG5L0dh
```
turboMaCk: I see both `yarn.lock` and `package-lock.json`. So which one is it?

src/registerServiceWorker.js
axypas: This is part of create-react-app, a toolchain to create React app quickly and without to have to deal with Webpack (a JS module bundler). It is the reference in the React community, by default the service worker is not enabled, you are right, I should have removed it.

turboMaCk: thanks for confirmation

package.json
```diff
@@ -0,0 +1,38 @@
...
+    "redux-thunk": "^2.3.0"
+  },
+  "devDependencies": {
+    "react-scripts": "3.1.1"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test --env=jsdom",
+    "eject": "react-scripts eject"
```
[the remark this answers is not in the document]
axypas: The main advantage of create-react-app is that it allows to update a lot of libraries automatically for you so that you can only focus on the business code and not in the toolchain. If for some reasons, you want to take control of everything, you can `eject` and then you will be responsible to update and configure what this toolchain was doing for you. So it is best to not eject, if possible.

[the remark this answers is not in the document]
turboMaCk: I understand this but I don't understand why is this part of scripts in `package.json`. This would be at max used just a single time so why not to call `react-scripts eject` directly. Let's compare the two things:

current:
- you add `eject` script to `package.json`
- you run it one day by `npm run eject`
- you remove it from `package.json` as it's no longer in used when project is already ejected

not doing this:
- one day you run `react-scripts eject`

Therefor I think this thing should not be there. Makes sense?

src/components/CatGallery.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import Gallery from "react-photo-gallery";
+
+const handlePhotoClick = handleClick => (event, { photo }) => {
+  handleClick(photo);
+};
```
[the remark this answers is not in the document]
axypas: Thanks for the comment, a library such as [lodash FP](https://github.com/lodash/lodash/wiki/FP-Guide) would help in this case.

src/modules/breeds/containers/Breeds.js
```diff
@@ -0,0 +1,85 @@
...
+import { fetchBreeds, fetchBreedsIfNeeded, selectBreed } from "../ducks";
+
+function mapStateToProps(state) {
+  const breedState = state.breedsReducer;
+  const { breeds, error, isLoading, selectedBreedId } = breedState;
+  return {
+    breeds,
+    error,
+    isLoading,
+    selectedBreed: breeds.find(breed => breed.id === selectedBreedId),
```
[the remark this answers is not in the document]
axypas: o(n)

src/modules/breeds/containers/Breeds.js
```diff
@@ -0,0 +1,85 @@
...
+    cats: breedState.cats.map(cat => ({ ...cat, src: cat.url }))
+  };
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+  return {
+    fetchBreeds: () => dispatch(fetchBreeds()),
+    fetchBreedsIfNeeded: () => dispatch(fetchBreedsIfNeeded()),
+    selectBreed: breed => dispatch(selectBreed(breed)),
```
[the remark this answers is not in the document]
axypas: Good point, I should have used `bindActionCreators` instead in this case in order not to have so many function compositions.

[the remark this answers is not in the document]
turboMaCk: Oh great. I didin't know `bindActionCreators` exist :joy:

src/modules/search/components/CatModal.js
axypas: The first check `breed` checks that the object has a truthy value, so not `null` neither `undefined`. Since the project is not using `null`, I could have written `breeds !== undefined && breeds.length > 0` to be more explicit, but since the project is not statically typed, I think the current approach is safer.

The React component does not make assumptions, so it does not know what will be passed as an array. If we were using statically typed languages, it would be indeed redundant because the contract would explicitly asked for an array.

src/modules/breeds/ducks.js
axypas: :) 

This is taken from https://github.com/erikras/ducks-modular-redux, basically, it put the action types, the actions and reducers in one file.

src/modules/search/components/CatModal.js
turboMaCk: > since the project is not statically typed, I think the current approach is safer.

is all I need tyo here. Even though I still it would quite a trouble if some truthy value other than array would be passed.

yarn.lock
axypas: `package-lock.json` is a mistake, thank you for catching that! Only `yarn.lock` should have been commited.

turboMaCk: no worries, this happens all the time.

src/modules/search/components/CatModal.js
axypas: Yes, static typing is handy in such cases :)

package.json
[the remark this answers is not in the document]
axypas: Your comment totally make sense now, I did not think of this. I guess it is just a shortcut over `yarn run react-scripts eject` to eject from the `react-scripts` version that is installed. If someone did `npx react-scripts eject`, the version of `react-scripts` and the latest version might create issues.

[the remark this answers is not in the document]
turboMaCk: that's a good point indeed. Anyway I would probably not be in favour of this.

src/components/CatGallery.js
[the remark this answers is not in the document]
turboMaCk: To be clear I would not use function composition for this in JS. in my opinion function like this are not much usefull unless defined as an infix operator. `compose(apply, extract)` is way too fetish (hey I know how composition works) compare to more practical haskell `apply . extract` or elm `apply << extract` or purescript `apply <<< extract`.

I'm usually even [against introducing them](https://github.com/NixOS/nixpkgs/pull/38422#issuecomment-533905258).

src/modules/breeds/containers/Breeds.js
```diff
@@ -0,0 +1,85 @@
...
+    unselectBreed: () => {
+      dispatch(selectBreed(undefined));
```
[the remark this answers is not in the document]
axypas: True, I always try to be very explicit.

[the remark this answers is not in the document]
turboMaCk: I've seen that in whole PR. Especially since you always explicitely state which properties of object to pass from function to function. I agree it's more maitainable than just passing giant object everywhere and losing control over what depends on what.

On the other hand I think types would be much better better solution. This seems almost like a workaround practice to me.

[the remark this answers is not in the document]
axypas: Upon reflection, I think I should have created an deselect breed method that would have made it more explicit.

src/modules/breeds/containers/Breeds.js
axypas: I don't really know to be honest, maybe I used the variable twice and the removed it and let it as is.

src/modules/breeds/containers/Breeds.js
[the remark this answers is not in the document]
turboMaCk: Well it's all about tradeoffs. It's hard to say what is better unless values are clearly stated. But I agree that it would be better if you value expliciveness more than expressiveness indeed. Btw you might want to try elm sometime - it's values system seems to be very much compatible with yours (even more than mine, I'm probably a bit more on expressive side of spectrum)

src/modules/breeds/containers/Breeds.js
turboMaCk: honestly I wouldn't even bother to comment this in regular review so ¯\_(ツ)_/¯

src/modules/breeds/ducks.js
axypas: You are totally right, it is a bit hackish.

turboMaCk:
Cool, once again thanks for all the time you've spend on this @axypas as well as all the discussion and for being so responsive to my comments. I think I have answers to all my questions :+1:

I'm passing this back to @gnikolaropoulos who will be in touch with you. Cheers!

axypas:
Thank you for your time @turboMaCk and for your extensive review!! Cheers!

README.md
renra: Same here.

PR #5, opened 2019-10-09
React Challenge

teliaz:


src/constants/models.ts
```diff
@@ -0,0 +1,96 @@
...
+        id: string
+        url: string
+    }
+}
+
+export interface IFavoriteBody {
+    image_id: string
+    sub_id: string
+}
+
```
gnikolaropoulos: I would personally have liked the models to leave closer to their modules. This file can get huge in a real world app. What do you think?

src/index.css
```diff
@@ -0,0 +1,44 @@
...
+  height: auto; 
+  width:auto;
+  margin: 0;
+  padding: 0;
+}
+
+.img-loading {
+  background: transparent url('/loader.gif') center no-repeat;
+}
+
```
gnikolaropoulos: What's your take on CSS in general? Would you go with pure css? Would you use any library? What about type safety on CSS?

src/store/actions/breed.ts
```diff
@@ -0,0 +1,20 @@
...
+    store: Store<AppState, AssociatedActions>,
+    breeds:  models.ApiListResult<models.ICatBreed>
+) => {
+
+    breeds = {...store.state.breeds};
+    breeds.status = models.StatusEnum.REQUESTED;
+    const rq = async () => {
+        const breedsRes = await apiBase('/v1/breeds');
+        breeds.data = breedsRes.data as models.ICatBreed[];
+        breeds.status = models.StatusEnum.SUCCESS;
```
gnikolaropoulos: what if the breedsRes request failed?

src/store/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    addFavorite: act.Favotite.addFavorite,
+    removeFavorite : act.Favotite.removeFavorite,
+    fetchCatsForBreed: act.Cat.fetchCatsForBreed,
+    fetchCat: act.Cat.fetchCat
+};
+
+
+export const useGlobal = globalHook<AppState, AssociatedActions>(React, initialState, stateActions);
+export default useGlobal;
+
```
gnikolaropoulos: I really don't like the index.ts name but I cannot find anything better right now! 😓

src/store/actions/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as Breed from './breed';
+import * as Cat from './cat';
+import * as Favotite from './favorite';
+
+export { Breed, Favotite, Cat }
```
gnikolaropoulos: What's the purpose of this file?

src/components/app/App.tsx
```diff
@@ -0,0 +1,54 @@
...
+              <Route path="/*" component={Page404} status={404} />
+            </Switch>
+          </ScreenClassProvider>
+        </ModalProvider>
+      </BrowserRouter>
+    </React.Fragment>
+  );
+}
+
+export default App;
```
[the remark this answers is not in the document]
teliaz: My approach was only based on the use of hooks. Finishing projects requirements saw that all components (serving single or multiple purposes) were pretty short so didn't bother to split it into modules.

Pros 
- Increases the ease of re-usability 
- Embraces the "do one thing and do it right"
- UX/UI front-end developer can pick up easily and re form the application 
- Easier tests and maintenance
- Separation of concerns

Cons
- Increased complexity
- Repetitive functionality implementation 
- Increased time for code navigation

src/components/breeds/Breeds.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const Breeds: React.FC = () => {
+
+  const [breeds, aBreeds] = useGlobal((state: AppState) => state.breeds, (actions: AssociatedActions) => actions);
+  const [selectedBreed, setSelectedBreed] = React.useState(initialBreed);
+  const [isOpen, setIsOpen] = React.useState(false);
+  React.useEffect(() => {
+    if (breeds.data.length === 0)
+      aBreeds.fetchBreeds();
+  });
```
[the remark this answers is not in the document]
teliaz: Still in Store though.

src/store/actions/breed.ts
teliaz: Sorry, I m glad that you spotted that. and yes , it is a big miss in my implementation. I should have implemented error handling in all responses.
There is the struct (ApiResult, ApiListResult) using generics  but a had to add error models to apply a clean approach.

src/store/actions/index.ts
teliaz: Thanks for asking. This would be ideal in models section.
I used it very heavily on other projects and is used as helper for imports in other components. 

With an import like this 
**import * as action from './../store/models**
u can have all actions with intellisense like 
action.[section]
**action.Cat.actionName1**
**action.Breed.actionName2**
etc etc

src/components/search/Search.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    ))}
+                </Row>
+                {imgs.length === 0 && <Spinner></Spinner>}
+            </Container>
+        </>
+    );
+};
+
+export default Search;
+
```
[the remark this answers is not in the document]
teliaz: Weird name on the app structure but still this could be easily extended to search for cat images,
so I named based on the API features. The challenge required only "search by breed"  and name stuck to Search.

src/store/index.ts
teliaz: Me neither. Naming can be frustrating at times, index is at least a generalized term.

src/components/breeds/Breeds.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+
+  function openImages(breed: ICatBreed) {
+    setSelectedBreed({ data: breed, status: StatusEnum.SUCCESS });
+    setIsOpen(true);
+  }
+  return (
+    <>
```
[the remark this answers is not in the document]
gnikolaropoulos: Probably it's the extension. I don't have that issue but I don't use this extension!

src/index.css
gnikolaropoulos: Apart from type safety, what CSS libraries would you advise? Sass? LESS? Anything else? Just plain old css?

teliaz: Using an external UI component framework with type safe css kinda depletes the idea on using a pre-compiler. All custom values are entered in a specific file, for custom theme or custom stylesheet.

Apart from using a external UI framework, I would prefer to use the component scope css for positioning and using a global scope css with a precompiler for support of custom/user-defined variables.

For basic stuff I would implement using  [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/components/app/App.test.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import App from './App';
+
+it('renders without crashing', () => {
+  const div = document.createElement('div');
+  ReactDOM.render(<App />, div);
+  ReactDOM.unmountComponentAtNode(div);
+});
```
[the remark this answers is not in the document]
teliaz: **Unit tests**
Using jest i would measure the testing coverage. Unit tests can provide a better development experience. 
Unit tests work better on pure components though. 
Using type safe typescript unit tests are not obligatory, especially those required when passing props or objects. 


**Integration Tests** 
Ideally a TDD Approach could be beneficial on integration testing providing component elements with custom test attributes. 
On integration testing sometimes a simple mock api can provide valuable information on applications breaks on external calls. 


**E2E tests** 
Being the outer layer of testing these tests can help to catch UI inconsistencies, authorization issues. As an external tool Katalon Studio would be a nice approach 
since it is agnostic of the underlying technology and can be assigned to people not knowing the internals of the application. 
Another solution would be cypress with the help of coveralls for coverage discovery, also the use of cypress fixture when mocking data. 

Things I would test and not test is based on the team approach. 
My personal opinion is that good ideas on integration tests can substitute tedious unit tests. 
 

A good test suite is based on the deployment schema. For developers a good test suite would provide fast, scoped tests for developing
and slow E2E tests should applied on a git hook for avoiding pushing them to master.

PR #6, opened 2019-10-14
vissarion - cat-lover-challenge

vissarionas:


gnikolaropoulos:
Hey @vissarionas! Thanks a lot for your time and effort, really appreciate it! 💪 

Just a couple of comments by just running your app, before I start a more proper review:
1) There was no need to combine the 3 views but I like your approach
2) You could have created a key on your own and just have that in the env file
3) Not renaming the env file (to let's say .env or .env.local etc) and let it as 'env.example' leaves the app in a really weird state! I can load images and navigate to breeds but favourites mechanism breaks completely
4) The fact that you combined the views leads to loading the same images again and again when I choose a breed
5) We don't want you to keep the cdn link to the photo. We want to have a link that when you open a new browser or a new tab it will get you to the app with the correct modal opened.

I will come back with a proper code review but I think you have something to work until then!

vissarionas:
Good morning , and really thanks for the comments so far!

2) I just don't like having api keys hardcoded and exposed in the repo.
3) I put some instructions in the README but probably they are not really easy to spot.
4) I totally missed that, but I could prevent this by checking the current breed in the store, and not triggering the api call, and so on.
5) This part with the link can be resolved either by drastically refactoring the app in a 'router' manner, or by providing the link of the app + url params (imageId), that will be consumed in some component mounting and accordingly update the store for enabling the modal. 

Thanks again!

gnikolaropoulos:
About point 3, I have seen the instructions and they are fine actually. It's strange though that you can get the app semi-working with just adding the key there and not renaming the file.

As for point 5, are you going to provide a fix or should we continue with the pull request?

vissarionas:
You should continue with the code review. Starting tomorrow I will be visiting Athens, so my time is not really in favour.

src/components/favourites/favourites.css
```diff
@@ -0,0 +1,14 @@
+#favourites-title {
```
renra: Hey @vissarionas Could you describe how you would structure stylesheets in a larger project?

src/components/breeds/breeds.test.jsx
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-undef */
+import React from 'react';
+import { cleanup, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import renderWithRedux from '../../redux/testRenderer';
+import Breeds from './breeds';
+import Api from '../../api';
+import mockResponses from '../../api/__mocks__/mockResponses';
+
+Api.getBreeds = jest.fn(() => mockResponses.getBreedsResponse);
```
renra: This seems to possibly affect other tests as well. Can you think of a way to limit the impact?

src/components/gallery/galleryComponents/tile/tile.jsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import { PropTypes } from 'prop-types';
+import './tile.css';
+
+const Tile = ({ url, onClick }) => {
+  const [visible, setVisibility] = useState(false);
```
renra: How would you do this if you were not allowed to use component local state?

src/components/modal/modal.jsx
```diff
@@ -0,0 +1,51 @@
...
+  const handler = (event) => {
+    if (event.keyCode === 27) {
+      callback();
+    }
+    document.removeEventListener('keydown', handler);
+  };
+  document.addEventListener('keydown', handler);
+};
+
+const Modal = () => {
```
renra: How would you go about creating a generic `Modal` component that others can use to render anything they want?

src/components/modal/modalComponents/footer/footer.jsx
```diff
@@ -0,0 +1,37 @@
...
+        url={url}
+        favouriteId={favouriteId}
+        toggleFavourite={toggleFavourite}
+      />
+    </div>
+  );
+};
+
+Footer.propTypes = {
+  modalProps: PropTypes.objectOf(PropTypes.oneOfType([
```
renra: Can you think of a more strict type-checking?

src/components/favourites/favourites.css
vissarionas: My approach so far is trying to keep CSS on component level, unless there is a clear need for some truly global CSS rule.

src/components/modal/modalComponents/footer/footerComponents/modalActions/modalActions.jsx
```diff
@@ -0,0 +1,62 @@
...
+);
+
+modalActions.propTypes = {
+  id: PropTypes.string,
+  url: PropTypes.string,
+  favouriteId: PropTypes.number,
+  toggleFavourite: PropTypes.func.isRequired,
+};
+
+modalActions.defaultProps = { id: '', url: '', favouriteId: null };
```
renra: What makes you choose the `''` or `null` as the default value?

src/components/breeds/breeds.test.jsx
vissarionas: I would do API.getBreeds = jest.fn(), and use mockImplementationOnce() in every test, with different responses. Or just use the default mocks from __mocks__ directory, and again mockImplementationOnce(), only on selected tests.

src/index.css
```diff
@@ -0,0 +1,13 @@
+body {
```
renra: What would you do to make common webpage elements look similar across browsers?

renra: I hope I'm not being too vague, feel free to ask for a hint :smile:

src/components/breeds/breeds.jsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  render() {
+    const { breeds, updateSearchParams } = this.props;
+    const allBreeds = { id: 'all', name: 'ALL' };
+    if (breeds && breeds.length) {
+      return (
+        <div data-testid="breeds">
+          <p id="breeds-title">BREEDS FILTER</p>
+          {[allBreeds, ...breeds].map((breed) => (
```
renra: How would you communitcate to the user that some network IO operation is going on or that some request important for this view has failed?

src/components/gallery/galleryComponents/tile/tile.jsx
vissarionas: I do not have a direct answer, but what comes first in my mind is ref, but again in a hooks manner.

src/components/modal/modal.jsx
vissarionas: I would use the props.children from this component while wrapping whatever other component I wanted with this, or I would make it more functional, accept component as argument, and return it 'decorated'. I think you can call it a HOC.

src/components/modal/modalComponents/footer/footerComponents/modalActions/modalActions.jsx
vissarionas: Ugly leftover. Just missed it. I think in this situation you can go with both.

src/index.css
vissarionas: I would say some CSS reset rules.

src/components/breeds/breeds.jsx
vissarionas: With a nice component instead of the null, explaining what happened in a human way, and give her/him the change to retry manually.

src/components/modal/modalComponents/footer/footer.jsx
renra: That would be cool. PureScript is another option although I've never used it myself. There's another, stricter option just with PropTypes though.

src/components/modal/modalComponents/footer/footerComponents/modalActions/modalActions.jsx
renra: I tend to use empty strings for variables that I pass as values to inputs.

src/components/breeds/breeds.jsx
renra: And how would you handle the state of the request?

src/api/__mocks__/index.js
```diff
@@ -0,0 +1,11 @@
+import mockResponses from './mockResponses';
+
+const mock = {
```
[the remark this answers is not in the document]
vissarionas: It is used in gallery.test.jsx. I mocked gallery tests this way just to showcase this approach.

src/components/breeds/breeds.test.jsx
renra: The second option looks better to me so that the API would be safely mocked away in all tests. For making sure I don't affect other tests I've come to use [sinon](https://sinonjs.org/)'s `sandbox` and `stub` and then reset them after each test.

src/components/favourites/favourites.css
renra: Are you used to some techniques of css pre- or post-processing?

src/components/gallery/galleryComponents/tile/tile.jsx
renra: Let's say the tile's visibility should be controllable by some other component - say on a button click. How would the flow look like throughout the app?

src/components/favourites/favourites.css
vissarionas: Not really. I just experimented with sass.

src/components/breeds/breeds.jsx
vissarionas: By catching the error in the saga, dispatching some fetch-error action, that updates the redux state, and reflect these changes in the specific component.

src/App.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+    fireEvent.click(tile);
+    expect(modal).toHaveAttribute('style', 'display: flex;');
+    fireEvent.click(modalFavouriteIcon);
+    expect(Api.saveFavourite).toHaveBeenCalledTimes(1);
+    fireEvent.click(modalCloseButton);
+    expect(modal).toHaveAttribute('style', 'display: none;');
+    fireEvent.click(breedButton);
+    expect(Api.getCats).toHaveBeenCalledTimes(2);
+  });
+});
```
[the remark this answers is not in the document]
vissarionas: Testing in general is a crazy good practice, and my first thoughts are coding confidence and code documentation. Up to now I am familiar with unit and integration testing, although sometimes these two are not really different theories.
About e2e, I have an understanding about selenium, but not something to work on yet.

src/components/gallery/galleryComponents/tile/tile.jsx
vissarionas: So both tile and button are sitting in the same parent component.
The button will pass the click handling to the parent, and the parent will re-render the tile with an updated prop that defines the style of it.

src/components/modal/modalComponents/footer/footer.jsx
vissarionas: Flow is the thing that I will try for sure.

src/components/gallery/galleryComponents/tile/tile.jsx
renra: And what if those two components could not have a direct common parent?

src/App.test.jsx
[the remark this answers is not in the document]
renra: In fact everything is a unit test. The only difference is how big or small the unit is :smile:

src/components/modal/modalComponents/footer/footer.jsx
renra: Here I was aiming at `PropTypes.shape` which is more strict than just `objectOf`. For example if I passed in an object which would have totally different keys, but values of the types you expect, `PropTypes` would be happy, but things could still fail.

renra:
Thanks @vissarionas for the time you invested into the coding and answering our questions. I think it's a sound implementation :+1:

src/components/gallery/galleryComponents/tile/tile.jsx
vissarionas: One approach would be by doing the update through redux state, or preferably I would use refs by exporting the tile component with forwardRef to access it elsewhere.

src/components/modal/modalComponents/footer/footer.jsx
vissarionas: I missed the importance of shape. Really good to know!

PR #9, opened 2021-07-09
Panayiotis Coding Challenge

pankgeorg:
# Hello!! 👋🏽👋🏽

Hope this finds you well :)

Find here an approach to challenge. It took something like 6 solid hours. It was nice! but also kind of confusing sometimes (kind-of-inception of cats in cats in cats).

I think I got most stuff in. In any case, do let me know what you think!

Contact me for questions/fuzziness clarifications.

Thanks!


gnikolaropoulos:
Hi @pankgeorg ! Thanks a lot for your effort! @teliaz and @ngkamperlo will do some comments of your code, I might also check a few things, so please wait a couple of days before a proper review! Again many thanks for your work, and I am glad you liked the challenge.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-prettier": "^3.2.0",
+    "eslint-plugin-react": "^7.21.5",
+    "prettier": "^2.2.1",
+    "prettier-stylelint": "^0.4.2",
+    "stylelint": "^13.9.0",
+    "stylelint-config-property-sort-order-smacss": "^6.4.0",
+    "stylelint-config-standard": "^20.0.0",
+    "stylelint-order": "^4.1.0",
+    "typescript": "^4.1.3"
```
ngkamperlo, line 35: Since typescript is added in the project, how come you chose to use plain JS?

src/App.js
```diff
@@ -0,0 +1,83 @@
...
+      >
+        <Router>
+          <Header />
+          <Switch>
+            <Route
+              exact
+              path={[
+                "/breeds",
+                "/breeds/breed/:breedinfoid",
+                "/breeds/cats/:breedid"
```
ngkamperlo, line 57: Here you could use camelcase maybe, WDYT?

src/components/BooleanField.jsx
```diff
@@ -0,0 +1,26 @@
...
+const useStyles = makeStyles((theme) => ({
+  root: {
+    display: "flex",
+    flexFlow: "column nowrap",
+    margin: theme.spacing(1)
+  }
+}));
+export default function BooleanField({ title, value }) {
+  const classes = useStyles();
+  if (value === null || value === undefined) return null;
```
ngkamperlo, line 15: Here you could use a util function, WDYT?

src/components/BooleanField.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className={classes.root}>
```
ngkamperlo, line 17: Since you have added material, you could also have used a flex container i.e Grid, how come you didn't use Grid instead?

src/components/BreedFetcher.jsx
```diff
@@ -0,0 +1,20 @@
...
+  const { breedinfoid: breedId } = useParams();
+
+  const [loading, cats = [], error] = useAPI(
+    ENDPOINTS.GET_ONE_BREED({
+      breed_id: breedId
+    })
+  );
+  const breed = cats?.[0]?.breeds?.find(({ id }) => id === breedId);
+  if (loading) return <Progress />;
+  if (!breedId || error || !breed) return null;
```
ngkamperlo, line 18: Another approach you could take was to return a React fragment instead of null. What benefits has each approach?

src/components/Error.jsx
```diff
@@ -0,0 +1,35 @@
...
+  children = (
+    <span>
+      An error occured :( <br /> We&apos;ll be back sooon
+    </span>
+  )
+}) {
+  const classes = useStyles();
+  useEffect(() => {
+    // Replace this with Centralized Error reporting: e.g. Stackdriver
+    window.console.error(error);
```
ngkamperlo, line 28: I think window is not needed here - WDYT?

src/components/Favourite.jsx
```diff
@@ -0,0 +1,72 @@
...
+    );
+  return (
+    <Fab
+      className={className}
+      onClick={() => favourite().then(refetch)}
+      color="primary"
+      aria-label="more"
+      variant="extended"
+    >
+      <FavoriteBorderIcon style={{ fill: "darkgrey", marginRight: 4 }} />
```
ngkamperlo, line 68: Why did you use style property in this elements?

src/components/Header.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+import AppBar from "@material-ui/core/AppBar";
+import Toolbar from "@material-ui/core/Toolbar";
+
+import { Link, useHistory } from "react-router-dom";
+
+const useStyles = makeStyles((theme) => ({
+  appBar: {
+    flexGrow: 1,
+    dispay: "flex",
```
ngkamperlo, line 12: Small typo here

src/pages/CatsList.jsx
```diff
@@ -0,0 +1,139 @@
...
+}));
+export default function CatsList({ page = 0 }) {
+  const classes = useStyles();
+  const { breedid: breedId } = useParams();
+  const [showMoreClicked, onClickShowMore] = useState(false);
+  const [loading, data, error] = useAPI(
+    ENDPOINTS.GET_IMAGE_SEARCH({
+      limit: PAGE_SIZE_INITIAL,
+      page,
+      order: "ASC",
```
ngkamperlo, line 55: This could be added as a constant too , WDYT?

src/pages/CatsList.jsx
```diff
@@ -0,0 +1,139 @@
...
+    return () => clearTimeout(timeout);
+  }, [loading]);
+  const memoizedCats = useMemo(
+    // We memoize this because otherwise it flickers
+    () =>
+      data?.map(
+        (cat) => ({ ...cat, cols: Math.round(Math.random() * 4) } || [])
+      ),
+    [data]
+  );
```
ngkamperlo, line 75: Could you use another method to prevent flickering?

src/pages/CatsList.jsx
```diff
@@ -0,0 +1,139 @@
...
+              color="primary"
+              aria-label="more"
+              variant="extended"
+            >
+              <AddIcon style={{ fill: "darkblue" }} />
+              More 😻😽
+            </Fab>
+          </div>
+        )}
+        <div className={classes.imageRoot}>
```
ngkamperlo, line 96: How could you reduce the size of this large component?

ngkamperlo (commented):
Very nice work Panayiotis! Thank you so much for your time mate! can you please check my comments when you have some time? Thanks in advance and again thanks for your effort!

teliaz:
Nice PR Panayiotis! Thanks for your time and effort!  

Some comments on your PR. 
- Have you got any experience using typescript in any fe framework? 
- It seems I got some console error while testing it locally. What do you this is the impact of such errors on a fe application? 
- Have you used any tool to log these? 
- Which tools/frameworks would you use to test a fe application? 

Again thanks for your submission!

src/components/BreedTemplate.jsx
```diff
@@ -0,0 +1,84 @@
+import { Card, CardContent, Typography } from "@material-ui/core";
+import { makeStyles } from "@material-ui/core/styles";
+import Display from "./Display";
+import BooleanField from "./BooleanField";
+import StarRating from "./StarRating";
+
+const useStyles = makeStyles((theme) => ({
```
teliaz, line 7: Would you consider breaking styles to a separate file?

src/components/Header.jsx
```diff
@@ -0,0 +1,49 @@
...
+    flexFlow: "row nowrap",
+    justifyContent: "flex-start",
+    alignItems: "center"
+  },
+  menuButton: {
+    marginRight: theme.spacing(2)
+  },
+  logo: {
+    height: 48,
+    borderRadius: "12px 0 12px 0",
```
teliaz, line 22: You could use another value here? WDYT?

src/hooks/useData.js
```diff
@@ -0,0 +1,45 @@
+import { useGet, useMutate } from "restful-react";
+import { ENDPOINTS } from "../constants";
```
teliaz, line 2: All things consider look good for that application scale. How would you structure this in a larger application?

package.json
pankgeorg: I like to add complexity as needed. Also, to be honest I'm not a typescript guru. In the projects I currently manage, we identify files to be `typescript`-ed 
- if we come to find more than 'a few' bugs in the logic
- if we really depend on the component (it's used multiple times in the code), so it's scary to leave untyped.
- (other red flags I can't think now).

src/App.js
pankgeorg: Yes. CamelCase would actually be a better choice here. I think you're right!

src/components/BooleanField.jsx
pankgeorg: Also `!=` but I like being explicit (PEP-8 way). The number of times I've missed a `false` or a `0` in these type of components is too high. This would be the first typescript refactor target, I guess, by my criteria above.

src/components/BooleanField.jsx
pankgeorg: Yes, `<Grid />` comes with a lot of functionality out of the box, and to be honest, it's scary to carry it around for cases like this. I just needed a column, so boom. (Also, it's not really a `Grid`, semantically, not that it this means anything 😇😊)

src/components/BreedFetcher.jsx
pankgeorg: I haven't used fragments (`<></>`) where I would return `null`, to be honest. `null` unmounts the node from the DOM completely, so that's nice. I'm using `Fragment`s to avoid the injection of unecessary or unwanted **container** nodes, as the [tutorial says](https://reactjs.org/docs/fragments.html). Also that fits better _my_ mental model of a `null` and a `Fragment` (which is very subjective, of course).

src/components/Favourite.jsx
pankgeorg: That's because I didn't have time to create color classes for my SVGs. Generally that's bad, and I do avoid it, even though I **_would_ prefer** the `style` / `fill/color` combo to a badly named class here, if it was alone (without the margin), as I find it more explicit.

src/components/Header.jsx
pankgeorg: 🤦🏽🤦🏽‍♂️🤦🏽‍♀️

src/pages/CatsList.jsx
pankgeorg: Yes, we could have it in `constants/index.js` with the rest of the constants the API uses. I really think of ASC/DESC and the rest of SQL keywords as constants in my mind so that didn't help in factoring this out early.

src/pages/CatsList.jsx
pankgeorg: To prevent _this_ type of flickering (which is, randomly changing columns) I could give static column size to all of the cats, but that would be boring 🐈 😊

More technically, I'll need some clarification - which type of flickering you have in mind?

src/pages/CatsList.jsx
pankgeorg: Lines 83-95 can be factored into
```jsx
const FabMore = ({show}) => show && <div className={classes.fabBar}>
            <Fab
              onClick={() => onClickShowMore(true)}
              color="primary"
              aria-label="more"
              variant="extended"
            >
              <AddIcon style={{ fill: "darkblue" }} />
              More 😻😽
            </Fab>
          </div>


```

pankgeorg: which would make them as `<FabMore show={!showMoreClicked} />`

pankgeorg:
> Nice PR Panayiotis! Thanks for your time and effort!

😊😇🙂

> Some comments on your PR.
> 
> * Have you got any experience using typescript in any fe framework?

Small experience with typescript with React, mostly my colleagues use it. We prioritize using Typescript to components that are used a lot or are changed a lot into bug-related PRs.

> * It seems I got some console error while testing it locally. What do you this is the impact of such errors on a fe application?

Can you clarify on this? Was it an error on the browser console or the npm command line?

`src/components/Error.jsx` uses `window.console.error` when rendered and the comment above it reads `    // Replace this with Centralized Error reporting: e.g. Stackdriver`
We use stackdriver in our FE tools to triage bugs and even to track them before they are being reported from users (without such a tool the errors are just scary for the poor user opening the console)

Please followup on this because I'm pretty sure I didn't understand correctly the question

> * Have you used any tool to log these?

Stackdriver with Google Cloud Platform's `Log Explorer`.

> * Which tools/frameworks would you use to test a fe application?

I'm using `jest` and `puppeteer`. Recently experimenting with Cypress but don't really know how to use it yet.

> Again thanks for your submission!

Thank you for your time reviewing!!

src/components/Header.jsx
pankgeorg: yes, `12px 0px` would be the same semantically, but I like being explicit!

src/components/BreedTemplate.jsx
pankgeorg: hmmm, that's a good question. I like the separation of concerns, but I also like readability. In that scale, (~10 LOCS) I like the styles being close to the code they're used.

If they get bigger, it's a question of why did they get bigger? Why did we divert from the theme so much? Do we have something really unique as a design? Do we need a new component?

When the answer to the questions points in the direction of 

> yes we need all these customizations

it's best to make a folder with the component and split the styles into a `styles.js` file.

src/components/Error.jsx
pankgeorg: Somehow adding a `window.` disables the eslint rule (any rule!). Yes it's not needed!

src/hooks/useData.js
pankgeorg: I think the ideal solution is to have some kind of API introspection, where an endpoint of the API would communicate the available operations/constants/endpoints/keywords and stuff. Then, it's all abiding to that 'protocol'. I'd love to discuss more and more specifically!

pankgeorg:
@teliaz, @ngkamperlo Thank you very much for your time, people!

Hope I answered all your questions, but do let me know if I missed something (possible) or if I wasn't clear/understandable/did not get your question (possible^2).

Cheers,
Panagiotis

package.json
ngkamperlo: Well to be honest the added complexity is only at the start of the project and then you are good to go. 
Are you using prop types instead of Typescript?

src/components/BooleanField.jsx
ngkamperlo: At the end of the day it is rendered in DOM as a simple div with the proper classes attached so I don't think of it as scary.
But what works for you best ;)

src/pages/CatsList.jsx
ngkamperlo: I think you could use in this case flex column pattern instead of manually calculating the width of each column. I think this is also the reason which creates issues in the UI such as below
image

package.json
pankgeorg: No I don't! My personal style is plain old javascript for most of components and some typescript for some selected few! I think it's important to state here that flexible though - above personal style is team consensus!

src/pages/CatsList.jsx
pankgeorg: Oh I see. Note that I'm not actively doing anything with widths and stuff; I wanted it to look strange (weird choice retrospectively). If I was to go for a more pacted view, I'd create numbers that add up to four: `[3, 1, 2, 2, 1, 3][i % 6]` instead of a random in `[1-4]` for example

Line 72 to `        (cat, i) => ({ ...cat, cols: [3, 1, 2, 2, 1, 3][i % 6]) } || [])`

I guess it was late and my creative self was not in its best artistic mode!

ngkamperlo: I think you should have handled it in a different way, e.g: create a grid container with children that take 33.3% or 25% of the total width of the row as this UI seems pretty odd. Also the width of the title seems completely wrong IMHO

pankgeorg: Yes, you're right. [ImageList](https://material-ui.com/api/image-list/#imagelist-api) does use the flex container internally with percentage of total width, but I didn't use it properly to get the right result. Furthermore, in your example, the cat image does not grow to get all space _its container has_. I should have handled it better.

pankgeorg:
@gnikolaropoulos, @teliaz, @ngkamperlo
Thank you very much for your time, people! I really appreciate :) 

Εις το επανιδειν 🍻🍸🫖👋🏽🙋🏽‍♂️🙋🏽‍♀️!!
Panagiotis

PR #10, opened 2021-07-23
one commit to rule them all

dioannou30:
Pull request for the cat-api challenge as requested.


src/App.css
```diff
@@ -0,0 +1,26 @@
...
+  color: black !important;
+}
+
+html * {
+  font-family: "Segoe UI", Tahoma, Geneva, Verdana, sans-serif;
+}
+
+.breed-image {
+  height: 15rem;
+  width: 15rem;
```
ngkamperlo, line 12: How come you used `rem` at this point?

src/App.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.details-table th {
+  text-align: end;
+}
+
+.alert-notification {
+  position: fixed;
+  top: 1em;
+  right: 1em;
```
ngkamperlo, line 22: Why did use `em` at this point? how come you didn't use `rem` as above?

src/App.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { CacheKeys } from "./constants/constants";
+import ErrorProvider from "./context/ErrorProvider";
+import DashboardLayout from "./layouts/DashboardLayout";
+import { uuid } from "./utilities/utils";
+
+const queryClient = new QueryClient();
+const subs_id = uuid();
+localStorage.setItem(CacheKeys.SUBS_ID, subs_id);
+
+function App() {
```
ngkamperlo, line 14: I see that in most places you have chosen `export const functionName = () =>` method. Why did you use `function` here?

src/components/AddToFavorites.tsx
```diff
@@ -0,0 +1,51 @@
...
+      addNotification({
+        message: "Cat added to favorites",
+        type: BootstrapVariants.SUCCESS,
+      });
+      setIsLoading(false);
+    } else {
+      addNotification({
+        message: DEFAULT_ERROR_MESSAGE,
+        type: BootstrapVariants.SUCCESS,
+      });
```
ngkamperlo, line 33: I think here a `setIsLoading(false)` was also needed, WDYT?

src/components/AddToFavorites.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <React.Fragment>
+      <Button
+        disabled={isLoading}
+        onClick={add}
+        variant="warning"
+        className="me-2"
+      >
+        <RiStarLine className="mb-1" />
```
ngkamperlo, line 45: Can you please describe what this component does?

src/components/Loading.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { Col, Row, Spinner } from "react-bootstrap";
+
+const Loading: React.FC = () => {
+  return (
+    <div>
```
ngkamperlo, line 6: Was this `div` element needed at this point?

src/pages/Home.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useInfiniteQuery } from "react-query";
+import Loading from "../components/Loading";
+import { CacheKeys } from "../constants/constants";
+
+const Home: React.FC = () => {
+  const {
+    data: cats,
+    status,
+    fetchNextPage,
+  } = useInfiniteQuery(CacheKeys.CatsList, fetchCats, {
```
ngkamperlo, line 12: Can you please explain what `useInfiniteQuery` does?

src/utilities/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+  );
+  return uuid;
+};
+
+export const extractImageFromUrl = (url?: string): string => {
+  if (url === undefined || url === null) return "";
+  const initialSplit = url.split("/");
+  const lastPart = initialSplit[initialSplit.length - 1];
+  const imageId = lastPart.split(".")[0];
```
ngkamperlo, line 21: I think the code here could be a bit more defensive, WDYT?

ngkamperlo (commented):
Great work mate! Thank you a lot for the effort you put in this PR! I have three asks from my part:
1. Can you please explain why did you use bootstrap? Are there any known good alternatives you are aware/heard of?
2. Can you please explain why did you use react-query? Are there any known good alternatives you are aware/heard of?
3. Can you please respond to the comments I made across the whole project? 
Again thanks for your time and the effort you put in this.

src/App.css
dioannou30: Given the fact that i don't assign anywhere font size, thus the whole project uses the default one, in this case _rem_ and _em_ are actually the same.

src/App.css
dioannou30: Again as above, no specific font-size set means that rem == em. For consistency i should have user "rem" here as well.

src/App.tsx
dioannou30: Its auto-generated from the create-react-app tool. It can work either way. Again for consistency maybe i should have made it a function expression.

src/components/AddToFavorites.tsx
dioannou30: You are correct this is a bug. ```setIsLoading(false)``` should go at the bottom outside of the ```else``` statement so in both cases the button becomes enabled again. 

Not only that, but the ```BootstrapVariant``` should be set on ```DANGER``` inside the else statement.

src/components/AddToFavorites.tsx
dioannou30: it handles the functionality of the "add to favorites" requirement, the button the user presses and the function to make the the api call when pressed.

src/components/Loading.tsx
dioannou30: It is not needed since everything can be inside a <Row /> element. I forgot to remove it after making it a bootstrap row. (it wasn't at first).

src/pages/Home.tsx
dioannou30: Its a react-query hook. It makes it easier to handle pagination. It helps to "load more" into existing data while handling caching (the main reason i used it).

src/utilities/utils.ts
dioannou30: You are right. In case url is null or undefined i make an unnecessary api call that returns an error. Fixed by raising a warning before making the call.

dioannou30:
> * Can you please explain why did you use bootstrap? Are there any known good alternatives you are aware/heard of?

Bootstrap is mostly a personal preference as i am used to the way it does things. We also use it on the project i currently work. A good alternative is React-Material. 

> * Can you please explain why did you use react-query? Are there any known good alternatives you are aware/heard of?

I wanted a way to cache the data fetched from the cat api plus handle the "load more" requirement . It alleviated the pain of writing my own cache handler. It always comes first upon searching for a way to cache an http call, thus it was the choise i preferred. I dont know any other libraries like react-query.

src/pages/Home.tsx
ngkamperlo: Great approach mate!

ngkamperlo:
Thanks for the effort mate! I have no other questions for you. Thanks a lot for your time!

PR #11, opened 2021-08-05
CatApi Project

bou8:


src/common/utils.js
```diff
@@ -0,0 +1,44 @@
...
+        headers: {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
+            'Access-Control-Allow-Origin': "*",
+            'x-api-key': API_KEY
+        },
+        body: JSON.stringify(param)
+    })
+        .then(res => res.json())
+        .catch(err => err);
```
ngkamperlo, line 31: How come some fetch requests are `async` and some are using `promise like` syntax?

src/reducers/reducers.js
```diff
@@ -0,0 +1,42 @@
...
+        case SET_ERROR:
+            return {...state, error: action.payload};
+        case SET_LOADING:
+            return {...state, isLoading: action.payload};
+        case SET_FAVOURITE_CAT:
+            return {...state, favouriteCat: action.payload};
+        case CLEAR_FAVOURITE_CAT:
+            delete state['favouriteCat'];
+            delete state['error'];
+            return state;
```
ngkamperlo, line 34: Can a bug be created by not destructuring the object here?

src/sagas/getCatImageSaga.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Worker saga to get the cat image. If succeeded set the cat image in redux state
+ * else sets the error in redux state
+ * {@link GET_CAT_IMAGE}
+ * @param action
+ */
+function* getCatImageSaga(action) {
+    try {
+        const image_id = action.payload;
```
ngkamperlo, line 18: How come you use both snake_case and camelCase?

src/styles/styles.js
```diff
@@ -0,0 +1,70 @@
+import {makeStyles} from "@material-ui/core";
```
ngkamperlo, line 1: Nice approach! are you aware of other css and element frameworks?

src/views/BreedImagesModal.js
```diff
@@ -0,0 +1,94 @@
...
+
+    return (
+        <Modal open className={classes.modal}>
+            <div className={classes.modalBodyBreedImage}>
+                <AppBar position="static" color="default">
+                <Tabs
+                    value={value}
+                    indicatorColor="primary"
+                    textColor="primary"
+                    onChange={(e, val) => setValue(val)}>
```
ngkamperlo, line 53: Is it better to write anonymous function as you did right here or create functions on function/class level ?

src/views/CatBreedsTab.js
```diff
@@ -0,0 +1,44 @@
...
+                </ImageListItem> : null
+            )}
+        </ImageList>
+        <Button disabled={isLoading.catBreeds}
+                variant="contained"
+                color="primary"
+                className={classes.btnLoadCats}
+                onClick={() => {
+                    getCatBreeds(page+1);
+                    setPage(page + 1)
```
ngkamperlo, line 35: Maybe this could be rewritten. Do you have any ideas?

src/views/CatImages.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Component for the 1st view.
+ * It contains an image list with a button for loading more cat images
+ */
+const CatImages = ({ catImages, isLoading, getCatImages, error }) => {
+    const classes = useStyles();
+    const [page, setPage] = useState(0);
+    useEffect(() => getCatImages(0),[getCatImages]);
+    return (
+        <React.Fragment>
```
ngkamperlo, line 19: What's the purpose of `React.Fragment?`

src/views/FavouriteImagesTab.js
```diff
@@ -0,0 +1,55 @@
...
+                </div> : null
+            ))}}
+        </ImageList>
+        <Button disabled={isLoading.favouriteImages}
+                variant="contained"
+                color="primary"
+                className={classes.btnLoadCats}
+                onClick={() => {
+                    getFavouriteImages(page + 1);
+                    setPage(page + 1);
```
ngkamperlo, line 47: I've seen this logic quite a few times. Do you think this could be refactored in some way?

ngkamperlo (commented):
Great work mate! Thank you a lot for the effort you put in this PR! I have three asks from my part:

Can you please explain why did you use material-ui? Are there any known good alternatives you are aware/heard of?
Can you please explain what typescript is? Have you used it? Are there any known good alternatives you are aware/heard of?
Can you please respond to the comments I made across the whole project?
Again thanks for your time and the effort you put in this.

ngkamperlo:
One last thing: I am pretty certain that something's problematic with the UI as for different cat ids we get the same picture. Can you please explain why this problem was caused?
image

src/views/FavouriteImagesTab.js
bou8: > One last thing: I am pretty certain that something's problematic with the UI as for different cat ids we get the same picture. Can you please explain why this problem was caused?
> image

It is not bug of the UI but some urls of the response show the same image.

src/views/CatImages.js
bou8: Fragments let you group a list of children without adding extra nodes to the DOM. So I prefer them instead of an empty 'div'.

src/views/BreedImagesModal.js
bou8: When a function has logic inside I prefer to write them on class level, but here I only set the state.

src/sagas/getCatImageSaga.js
bou8: I used the same name 'image_id' as the query param of the api. With that way I passed the param directly as {image_id} instead of {image_id: '...'}. Generally I prefer camelCase.

src/reducers/reducers.js
bou8: this action is called every time the user submits a favorite cat. I want these keys on state to be cleared in order to show the new messages from the response in the UI.

src/styles/styles.js
bou8: Thank you. Yes Bootstrap, Foundation, Skeleton, Primer

src/common/utils.js
bou8: My fault I forgot to remove here the async/await keywords

src/views/CatBreedsTab.js
bou8: I could use the 'useReducer' hook but I had only one state so I thought it was not so necessary.

src/views/FavouriteImagesTab.js
bou8: > I've seen this logic quite a few times. Do you think this could be refactored in some way?

replied above

src/reducers/reducers.js
ngkamperlo: Yes but a new object is not returned as it should ({... state}). Are you aware of the problems it can create in redux?

src/styles/styles.js
ngkamperlo: Have you used any of those professionally?

src/views/BreedImagesModal.js
ngkamperlo: I think anonymous functions declared in the return body, create a performance penalty in react, that's why I mentioned that. Are you familiar with it?

bou8:
> Great work mate! Thank you a lot for the effort you put in this PR! I have three asks from my part:
> 
> Can you please explain why did you use material-ui? Are there any known good alternatives you are aware/heard of?
> Can you please explain what typescript is? Have you used it? Are there any known good alternatives you are aware/heard of?
> Can you please respond to the comments I made across the whole project?
> Again thanks for your time and the effort you put in this.

Thank you.
- I used material-ui because I have experience to it and it has a lot of free elements. Also I have used Ant Design and PrimeNg.
- Yes I have used typescript with Angular and React also. It is type oriented and help catch mistakes early through a type system. Alternatives Javascript, Elm, Babel, Dart
- Replies to the comments below.

src/views/CatBreedsTab.js
ngkamperlo: No, no even simpler I was referring to page+1 code. It could be refactored to a variable and used for both sets

src/views/BreedImagesModal.js
bou8: For this reason all my components are functional instead of class to have better performance and avoid the continuous rendering.

src/styles/styles.js
bou8: Bootstrap and Primer

src/reducers/reducers.js
bou8: Right I forgot to put spread operator.

src/views/BreedImagesModal.js
ngkamperlo: I was referring to this issue right here: https://stackoverflow.com/a/48740930/1485965

src/reducers/reducers.js
ngkamperlo: If you do not put the spread operator or do not return a new object in general its usual that redux will not detect the changes and will not run the reducer function.

src/common/constants.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Constant unique for the user. Used to get and submit the user's favourite images
+ * {@see getFavouriteImagesSaga @link submitFavouriteCatSaga}
+ */
+export const MY_USER_ID = 'bou8Cat';
+
+/**
+ * Constant unique for the user. The api key is used to the request headers apis.
+ */
+export const API_KEY = '43abbc90-acfe-4c51-955b-3258248e5f8d';
```
teliaz, line 10: I find that very helpful so we can actually test your test with minimal setup. But can think of a way that you could hide sensitive information?

src/index.js
```diff
@@ -0,0 +1,38 @@
...
+import watchers from "./sagas/watchers";
+import {BrowserRouter, Route} from "react-router-dom";
+import CatImageModal from "./views/CatImageModal";
+import BreedImagesModal from "./views/BreedImagesModal";
+
+// create the saga middleware
+const sagaMiddleware = createSagaMiddleware();
+
+// mount it on the Store
+const store = createStore(
```
teliaz, line 19: Redux is always a nice choice! 
Can you explain its strengths and drawbacks against other state management solutions?

src/reducers/reducers.js
teliaz: Following this comment have you used any library that ensure state immutability?

src/sagas/deleteFavouriteCatSaga.js
```diff
@@ -0,0 +1,26 @@
+import {call, put} from 'redux-saga/effects'
```
teliaz, line 1: Have you used redux-thunk?  If so what are the drawbacks of each one?

src/styles/styles.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+    modalBodyCatImage: {
+        backgroundColor: theme.palette.background.paper,
+        border: '2px solid #000',
+        boxShadow: theme.shadows[5],
+        padding: theme.spacing(2, 4, 3),
+        width:'600px',
+        maxHeight: '500px',
+        overflow: 'auto',
+        fontSize: '14px'
```
teliaz, line 31: I see a mixed usage of values as string and numbers. Can you change any of those to make this simpler? 

My reasoning here is that when passing component props in makeStyles, I would prefer to have simpler ternary operators instead of string template literals.

teliaz:
Very nice work @bou8! Thanks for submitting code for our challenge! 

Following @ngkamperlo, can you respond to my comments?

src/reducers/reducers.js
bou8: yes 'immutable.js' and 'immutability helper'

src/common/constants.js
bou8: Secret data can be stored in backend and if client wants them, it can be accessed by making request to the server.

src/index.js
bou8: Pros: 
- Central store, any component can access any state from the store
- Prevents unnecessary re-renders, as when the state changes it returns new state
- Persists the state of a component even after the component has unmounted
- Testing will be easy as UI and data management are separated
Cons: 
- Any component can access the data which can cause security issues
- The reducer updates the state by returning a new state every time which can cause excessive use of memory
- Boilerplate code

src/sagas/deleteFavouriteCatSaga.js
bou8: Yes I have used it.
Cons of redux-thunk: Action creators may hold too much async logic, More difficult to scale up and to test
Cons of redux-saga: More boilercode code and more difficult to understand than redux-thunk

src/styles/styles.js
bou8: yes instead of having different classes for  'modalBodyCatImage' and 'modalBodyBreedImage' , I could have only one and passing a prop in makeStyles with ternary operator for width and height.

bou8:
> Very nice work @bou8! Thanks for submitting code for our challenge!
> 
> Following @ngkamperlo, can you respond to my comments?

Thank you.

src/views/CatImageDetailsTab.js
```diff
@@ -0,0 +1,59 @@
...
+             alt={catImage.url}
+             height={250}
+             width={300}
+             className={classes.marginImg}
+        /><br/>
+        <Button variant="contained"
+                color="primary"
+                className={classes.btnLoadCats}
+                onClick={() => submitFavouriteCat(match.params.id)}>
+            FAVOURITE
```
padelis, line 32: If a cat is already favored, the button shouldn't have a different label?

padelis (commented):
How does every modal opens in a new tab? Isn't a bit strange user experience?
Also this approach doesn't beat the purpose of using something like redux? (I mean the same data are fetched again and again) WDYT?

src/views/FavouriteImagesTab.js
padelis: This is due to the request is always the same.
Something like `v1/images/search?limit=10&page=${page}&order=random` would solve this issue.

src/views/CatImageDetailsTab.js
bou8: > Nice work @bou8! Thank you for your time and effort!
> 
> How does every modal opens in a new tab? Isn't a bit strange user experience?
> Also this approach doesn't beat the purpose of using something like redux? (I mean the same data are fetched again and again) WDYT?

In the requirements said that user can navigate directly to the route so each view is independent from the other. For this reason modal opens in a new tab. When you are in the view and doing some actions like load, submit or delete data are not fetching again and again.

bou8: > If a cat is already favored, the button shouldn't have a different label?

For sure improvements can be done

padelis: Yes navigating directly to a view was in the requirements, but the end result wouldn't be better (clearly from a user's perspective) if the views where combined in one page?

src/views/CatImageDetailsTab.js
```diff
@@ -0,0 +1,59 @@
...
+                <b>Description: </b>{breed.description}<br/>
+                <b>Temperament: </b>{breed.temperament}<br/>
+                <b>Weight: </b>{breed.weight && breed.weight.imperial}<br/>
+                <b>Metric: </b>{breed.weight && breed.weight.metric}<br/>
+                <b>Life span: </b>{breed.life_span}<br/>
+                <b>Cfa: </b> <a href={breed.cfa_url} target="_blank" rel="noreferrer">{breed.cfa_url}</a><br/>
+                <b>Vet street: </b><a href={breed.vetstreet_url} target="_blank" rel="noreferrer">{breed.vetstreet_url}</a><br/>
+                <b>Vca hospitals: </b><a href={breed.vcahospitals_url} target="_blank" rel="noreferrer">{breed.vcahospitals_url}</a><br/>
+                <b>Wikipedia: </b><a href={breed.vcahospitals_url} target="_blank" rel="noreferrer">{breed.wikipedia_url}</a><br/>
+                <div className={classes.rating}><b>Adaptability: </b><Rating value={breed.adaptability} size="small"/></div>
```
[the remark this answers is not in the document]
bou8: yes waring in the console more testing was needed

src/views/CatImageDetailsTab.js
bou8: From my point view the modals should not open in a different route

src/views/FavouriteImagesTab.js
bou8: Request is not the same because in each load request for the next 10 images. You can see the urls in the response are different but they some are referred to the same image. Also in the documantation said: 
"NOTE: Pagination can only be performed when the order query parameter is Desc or Asc - Not the default Rand"

PR #13, opened 2021-09-20
Assignment solution submit

costaioannou:


src/app/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+  useEffect(() => {
+    const isNotInFavouritesPage = window.location.pathname !== '/favourites';
+
+    if (isNotInFavouritesPage) {
+      dispatch(fetchFavourites());
+    }
+  }, []);
+
+  return (
+    <Flex direction="column" h="100vh">
```
ngkamperlo, line 27: Why do you use `vh` here? can you please explain its benefits?

src/breeds/BreedsPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Link } from 'react-router-dom';
+import { useAppDispatch, useAppSelector } from '../store';
+import {
+  breedsAreLoading, fetchBreedsForPage, selectAllBreedsByPage, selectTotalBreeds,
+} from './breedsSlice';
+import { Breed } from '../common/models';
+import { Image, Loader } from '../common/components';
+import { Pagination } from './Pagination';
+
+const BreedCard = ({ breed }: {breed: Breed}): JSX.Element => (
```
ngkamperlo, line 14: Could have this been rewritten an a simpler way?

src/breeds/breedsSlice.ts
```diff
@@ -0,0 +1,95 @@
...
+  byId: {},
+  imagesById: {},
+  total: 0,
+  loading: false,
+  error: null,
+};
+
+const breedSchema = new schema.Entity('breeds');
+
+export const fetchBreedsForPage = createAsyncThunk(
```
ngkamperlo, line 28: Can you briefly explain what Thunk does?

src/modal/Modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      dispatch(addFavourite(imageId));
+    }
+  };
+
+  return (
+    <Button
+      colorScheme={isFavourite ? 'red' : 'blue'}
+      isLoading={isLoading}
+      onClick={() => toggleFavourite()}
```
ngkamperlo, line 62: Could this have been rewritten in a different way?

ngkamperlo, line 62: Could this have been rewritten in a different way?

src/reportWebVitals.ts
```diff
@@ -0,0 +1,17 @@
...
+      getCLS(onPerfEntry);
+      getFID(onPerfEntry);
+      getFCP(onPerfEntry);
+      getLCP(onPerfEntry);
+      getTTFB(onPerfEntry);
+    });
+  }
+};
+
+export default reportWebVitals;
```
ngkamperlo, line 17: Can you please explain the usage of this file?

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
```
ngkamperlo, line 12: Can you please explain what this does?

ngkamperlo (commented):
Very nice work mate! thank you very much for the time you took for the submission! Can you please also answer to us the following questions?
1. Why did you use Typescript?
2. Why did you use chakra-ui?
3. If you had a little bit more time, what you would have dome differently?
Again thanks for your time and wish you the best of luck on your submission! :)

src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const Constants = {
```
teliaz, line 1: Nice one! Comments are also very helpful here!

src/favourites/FavouritesPage.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  afterAll(() => server.close());
+
+  test('loads initial favourite images on load', async () => {
+    renderComponent(<FavouritesPage />);
+
+    expect(screen.getByText('Loading...')).toBeInTheDocument();
+  });
+
+  test('initially loads the favourite images', async () => {
```
teliaz, line 21: Thanks for taking the time to implement tests in your PR! 
How would you handle data related tests? Can you imagine a way this test would break?

src/favourites/FavouritesPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+} from '@chakra-ui/react';
+import { useAppDispatch, useAppSelector } from '../store';
+import {
+  Card, Empty, Image, Loader,
+} from '../common/components';
+import {
+  favouritesAreLoading, fetchFavourites, removeFavourite, selectAllFavourites,
+} from './favouritesSlice';
+
+export const FavouritesPage = (): JSX.Element => {
```
teliaz, line 13: Why are you chose to use `JSX.element` here? Are there any other alternatives and what are their benefits ?

src/mocks/handlers.ts
```diff
@@ -0,0 +1,75 @@
+import { rest } from 'msw';
+import { v4 as uuidv4 } from 'uuid';
+import { mockBreedsPage1, mockBreedsPage2, mockFavourites } from './responses';
+
+export const handlers = [
+  rest.get('https://api.thecatapi.com/v1/images/search',
```
teliaz, line 6: You could use the constant here? wdyt?

src/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Provider } from 'react-redux';
+import {
+  BrowserRouter as Router,
+} from 'react-router-dom';
+import reportWebVitals from './reportWebVitals';
+import { App } from './app';
+import { store } from './store';
+
+ReactDOM.render(
+  <React.StrictMode>
```
teliaz, line 14: What is this node used for?

teliaz (commented):
Nice PR @costaioannou ! Thank you for the effort you put into the challenge! 

Can you please respond to the comments I made across the whole project?

src/breeds/BreedsPage.tsx
costaioannou: You mean without destructuring the props ? Like this `const BreedCard = (breed: Breed)...` ?

src/breeds/breedsSlice.ts
costaioannou: It's a middleware that sits between actions and reducers. It receives actions, does some calculations (fetching data etc) and returns new actions

src/modal/Modal.tsx
costaioannou: `onClick={toggleFavourite}` damn missed that

costaioannou: `onClick={toggleFavourite}` damn missed that x2

src/reportWebVitals.ts
costaioannou: It helps to measure and analyze the performance of the app

tsconfig.json
costaioannou: It allows default imports from modules when they got no default exports

costaioannou (commented):
1. Why did you use Typescript?
For static type checking and better readability. Maybe for this assignment is an overkill regarding the size and complexity.

2. Why did you use chakra-ui?
It's a simple and solid UI library. I used a UI library mainly for productivity reasons

3. If you had a little bit more time, what you would have dome differently?
For sure better UX in error handling and refactor some components to be more reusable

src/favourites/FavouritesPage.tsx
costaioannou: I prefer `JSX.element` , despite that TS can infer it (eslint was pointing to add a return type :P)

I know that I can add the FC type that accepts generics but it has some pitfalls like 
```
React.FC<Props> defines an optional 'children' prop. 
This makes you component API non explicit, and will autocomplete for `children`, 
when you may not be handling the property
```

src/mocks/handlers.ts
costaioannou: You mean reuse the url from a constant ? Yes ofc

src/index.tsx
costaioannou: It's a tool that runs some checks that helps you write better components

src/favourites/FavouritesPage.test.tsx
costaioannou: When there will be no images?  In tests I am mocking the response so I am sure I will receive images. (If I havent answered your question, can you elaborate ? 😄 )

teliaz: I would expect something like this test is related to the current API token and is passable based on the fact that I have favourited images. Makes sense?

ngkamperlo:
Thanks a ton for the answers mate! and again thx for the effort you put into this !

costaioannou:
@ngkamperlo @teliaz Thank you for the review!

PR #15, opened 2021-09-28
Test project completed

mcAnastasiou:


src/components/common/ImagesListWrapper/ImagesListWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import styles from './ImagesListWrapper.module.scss';
+
+type Props = {
+  children: React.ReactNode;
+};
+
+const ImagesListWrapper: React.FC<Props> = ({ children }) => {
```
ngkamperlo, line 8: How come you didn't use `PropsWithChildren`?

src/components/common/Modal/Modal.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onClose: () => void;
+};
+
+const Modal: React.FC<Props> = ({ children, show, onClose }) => {
+  if (!show) {
+    return null;
+  }
+
+  const onModalContentClick = (event: React.MouseEvent) => {
+    event.stopPropagation();
```
ngkamperlo, line 16: Why did you use that here?

src/components/common/Modal/Modal.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  width: 100%; /* Full width */
+  height: 100%; /* Full height */
+  z-index: 100;
+  background-color: rgba(31, 32, 41, 0.75);
+}
+
+.modal {
+  position: fixed;
+  padding: 20px;
+  width: 60vw;
```
ngkamperlo, line 17: Why did you use `vw` here? Could you have used something else ?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
```
ngkamperlo, line 12: What does this prop do?

ngkamperlo (commented):
Thank you very much for the effort you put into this mate! can you please check my questions below along with the following ones:
1. How did you choose typescript for your project?
2. How come you used scss? 
3. How come you used redux? are there any other alternatives you have used?
Thanks again for your submission and waiting for your responses !!

src/App.module.scss
```diff
@@ -0,0 +1,4 @@
+.content {
```
teliaz, line 1: How did you chose to not use a library?  Are you familiar with any react ui libraries?

src/components/Breeds/Breeds.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const dispatch = useDispatch();
+  const [showModal, setShowModal] = useState(false);
+  const [selectedBreed, setSelectedBreed] = useState<BreedType>();
+  const breeds = useSelector(allBreedsSelector);
+  const breedsLoading = useSelector(breedsLoadingSelector);
+  const images = useSelector(allImagesSelector);
+
+  useEffect(() => {
+    dispatch(getBreeds());
+  }, [dispatch]);
```
teliaz, line 27: Why did you choose to include dispatch to the dependency array?

src/components/Home/Home.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { getImages } from 'redux/images/actions';
+import { allImagesSelector, imagesLoadingSelector } from 'redux/images/selectors';
+import { ImageType } from 'types/images';
+import styles from './Home.module.scss';
+import ModalContent from './ModalContent/ModalContent';
+
+const Home = () => {
+  const history = useHistory();
+  const dispatch = useDispatch();
+  const { id } = useParams<{ id?: string }>() ?? {};
```
teliaz, line 21: Can you explain why you used both optional and null coalescing?

src/types/images.ts
```diff
@@ -0,0 +1,20 @@
+import { BreedType } from './breeds';
+
+export type CategoriesType = {
```
teliaz, line 3: What is the difference between types and interfaces? Why would you choose one over the other?

teliaz (commented):
Thank you for taking the time to complete the test, very nice submission @mcAnastasiou ! 
Can you please check my comments?

src/components/Home/ModalContent/ModalContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const history = useHistory();
+
+  const onGoToBreedsClicked = () => {
+    history.push(Routes.breeds.index);
+  };
+
+  const renderBreed = () => {
+    if (breeds.length > 0) {
+      return breeds.map((breed) => {
+        return <BreedDetails name={breed.name} description={breed.description} onClick={onGoToBreedsClicked} />;
```
thsarig, line 37: missing here the `key` prop drives to React console errors about missing unique key prop.

src/constants/routes.ts
```diff
@@ -0,0 +1,11 @@
+export const Routes = {
```
thsarig, line 1: I like the usage of this centralized `Routes` values 💪

src/components/common/Header/Header.tsx
```diff
@@ -0,0 +1,22 @@
+import { Routes } from 'constants/routes';
+import useActiveHeader, { HeaderTabEnum } from 'hooks/useActiveHeader';
+import React from 'react';
+import { Link } from 'react-router-dom';
+import styles from './Header.module.scss';
+import HeaderTab from './HeaderTab/HeaderTab';
+
+const Header = () => {
+  const { activeTab } = useActiveHeader();
```
thsarig, line 9: What drives you to follow the approach of a custom hook here?

thsarig (commented):
Thanks a lot @mcAnastasiou for your effort and your time on this assignment.

src/components/common/ImagesListWrapper/ImagesListWrapper.tsx
mcAnastasiou: I was not aware of this. Good to know! Thank you

src/components/common/Modal/Modal.tsx
mcAnastasiou: I didn't want to close the modal when the user clicks the content and not the backdrop. If i don't stop the propagation, the event will be propagated and will be handled by the backdrop which is the parent

src/components/common/Modal/Modal.module.scss
mcAnastasiou: i used viewport width in order to be more responsive. I could use % as an alternative.

tsconfig.json
mcAnastasiou: AllowJs, allows js to be imported in ts and tsx files
skipLibCheck, skips the check of declaration files, which leads to faster compilation
esModuleInterop, typescript should not treat CommonJS/AMD/UMD as es6Modules when this feature is enabled
allowSynteticDefaultImport: 

import React from "react";
instead of:
import * as React from "react";

To be honest, this file was autogenerated by create react app so probably not all of them are needed in every case

src/App.module.scss
mcAnastasiou: I try to avoid adding libraries because if there are many, it is hard in a big project to keep them up to date. Also it reduces the size of the bundles. Although i focused on the architecture and in the react part and not in the css. If i was paying more attention to css, i would have used React material ui, because of the convenience.
I am aware of React material ui and React bootstrap. I have used the first one in a personal project.

src/components/Breeds/Breeds.tsx
mcAnastasiou: ESLint complains about leaving the array empty, because dispatch is a result of the useDispatch hook, which could change in rerenders. Although i have tested this and it doesn't change, so it could be similar to []. Personally i think all dependencies that are being used in a hook and could change their reference, they should be added in the dependency array.  The developer is responsible to handle the number of time the callback function will be called using a useRef()

src/components/Home/Home.tsx
mcAnastasiou: The id which is a path param, is not defined in all cases. If it is defined, open the modal otherwise, just display the result. I ussed the null coalescing, because useParams could return undefined and the destructuring will fail in that case

src/types/images.ts
mcAnastasiou: Types and interfaces are very similar. Almost all features of an interface are available in type. the key distinction is that a type cannot be re-opened (change) after it has been created

src/components/Home/ModalContent/ModalContent.tsx
mcAnastasiou: You are right. My mistake that i didn't check the console to see any logs

src/constants/routes.ts
mcAnastasiou: Thank you!

src/components/common/Header/Header.tsx
mcAnastasiou: I think it makes the code more readable. I moved the logic to choose the selected tab in a hook because i needed the useLocation and useState. So a function could not be used. In the specific case where the Header component is very simple i could move the implementation to the component, rather than the hook. But i have seen that many times a simple component could become more complex as the app grows, so i try to avoid a future refactoring

mcAnastasiou:
Thank you all for you time and the comments that you wrote!

PR #16, opened 2021-10-06
feat(code): finished test

aliakakis:
The final test code as requested for the test


aliakakis:
Thank you for taking the time to comment on my code :)
In the README.md there are reasons for using the various libraries. Please have a look.
Choosing JS was based on React using JS as the de facto scripting language. TypeScript in my opinion is not idiomatic JS and is an extra dependency. Personally, I can see TS used for typing ajax JSON based responses, but in general I have seen TS overused making the code difficult to read and reason about. Moreover, JS never did have support for concepts TS is providing e.g. generics, tuples etc. making it more attractive for C# developers. Proper code reviews and careful use of jsdocs can do miracles :)

PR #17, opened 2021-11-02
Complete functionality of the CatLover app

alexanderkont:


src/components/App.jsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }, [location])
+
+  if (
+    location.state && 
+    location.state.background &&
+    prevLocation !== location
+  ) {
+    isModalOpen = true
+  }
```
ngkamperlo, line 32: Could this have been rewritten so that a `const isModalOpen` have been used instead?

src/components/breeds/BreedListItem.styles.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+  breedName: {
+    position: "absolute",
+    bottom: 0,
+    left: 0,
+    background: "rgba(0, 0, 0, 0.7)",
+    color: "#fff",
+    width: "100%",
+    textAlign: "center",
+    font: "18px/25px Helvetica",
```
ngkamperlo, line 23: Could font have been set in a higher place for maybe the whole application?

src/components/breeds/BreedView.styles.js
```diff
@@ -0,0 +1,27 @@
...
+  row: {
+    margin: [0, 0, 10, 0]
+  },
+  attrs: {
+    display: "grid",
+    gridRowGap: "5px"
+  },
+  attribute: {
+    display: "grid",
+    gridTemplateColumns: "1fr 2fr",
```
ngkamperlo, line 22: Can you please explain the usage of `fr`?

src/components/images/BreedDetails.styles.js
```diff
@@ -0,0 +1,27 @@
+import { createUseStyles } from "react-jss"
+
+export const useBreedDetailsStyles = createUseStyles({
+  breedDetails: {
+    margin: [10, 0, 0],
+    font: "15px/18px Helvetica"
+  },
+  detailsRow: {
+    margin: [0, 0, 5, 0],
```
ngkamperlo, line 9: Could this have been rewritten since you only need to set one value?

ngkamperlo (commented):
Thank you for the effort you put into this mate! Can you please check my questions below and please answer also the following:
1. How come you did not use typescript?
2. How come you used jss? Have you used this in the past?
Again thanks for the time you put into this and looking forward to your answers :)

src/components/App.jsx
alexanderkont: Nice catch! That's the way to do it:
`const isModalOpen = (
    location.state && 
    location.state.background &&
    prevLocation !== location
  ) || true`

Had a different implementation and after changing to that, didn't thought of moving it to a constant.

src/components/breeds/BreedListItem.styles.js
alexanderkont: Definitely! I come from a product that we need to be able to override the font attributes per component, so it has become a habit by now.

src/components/breeds/BreedView.styles.js
alexanderkont: So, 'fr' stands for 'fraction', and on grid templates, is used to describe the area that each item will cover.
In that scenario, the first column will occupy the 1/3 of the page width and the second column the 2/3.

src/components/images/BreedDetails.styles.js
alexanderkont: Do you mean something like that: `marginLeft: "5px"` ?
Don't know of any significant difference between the two approaches to be honest, so I usually include all directions.

alexanderkont:
> Thank you for the effort you put into this mate! Can you please check my questions below and please answer also the following:
> 
> 1. How come you did not use typescript?
> 2. How come you used jss? Have you used this in the past?
>    Again thanks for the time you put into this and looking forward to your answers :)

src/components/images/BreedDetails.styles.js
ngkamperlo: This would probably be equal to `marginBottom: '5px'` . It's just cleaner-there is no significant difference

alexanderkont: Correct, `marginBottom` is the correct direction here, sorry. 
I definitely take your point about being clearer to you.

PR #20, opened 2022-02-16
First and Last commit

johnpaps13:
Thank you for your time for the code review. I am looking forward your feedback! :D

Please use npm install and npm start to run the app. Don't forget the readme file for more info.


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,17 @@
+import Card from "@mui/material/Card";
+import CardContent from "@mui/material/CardContent";
+
+const CardStyle = {
+  minWidth: 375,
+  maxWidth: 325,
+};
+
+export default function BasicCard({ item }: { item: any }) {
```
ngkamperlo, line 9: How come you use `any` instead of a concrete interface?

src/components/ImageList/ImageList.tsx
```diff
@@ -0,0 +1,93 @@
...
+import styled from "@emotion/styled";
+import mq from "../../styling/mediaQueries";
+import Snackbar from "../../components/SnackBar/Snackbar";
+import Card from "@mui/material/Card";
+import Modal from "../Modal/Modal";
+import CardMedia from "@mui/material/CardMedia";
+import { CardActionArea } from "@mui/material";
+
+interface ImageListProps {
+  catsData: Array<object>;
```
ngkamperlo, line 11: Why don't you describe what catsData exactly is about?

ngkamperlo (commented):
Hello mate,
Thank you very much for the effort you put into this one! Apart from the questions below can you please answer the following:
1. Why did you use typescript and you added so many `any` properties?
2. How come you went with emotion library?
Again thank you for the effort you put into this assignment!

johnpaps13:
> Hello mate, Thank you very much for the effort you put into this one! Apart from the questions below can you please answer the following:
> 
> 1. Why did you use typescript and you added so many `any` properties?
> 2. How come you went with emotion library?
>    Again thank you for the effort you put into this assignment!

Hello there and thank you for your time,
1. My experience with typescript is not great (about 3 months), as previously i was using propTypes. I mentioned in my readme that i wanted to remove the "any" that have been left if i had more time. Although, i think that typescript solves many problems of javascript in general and it is great practice especially in great scale
2. I am fan of css-in-js style and emotion is a library that i have used before professionally. Although i understand why many people prefer css modules, i find it a bit verbose. It is just a matter of preferance.

src/components/Card/Card.tsx
johnpaps13: I forgot it to put interface there :/ It was only one prop so i missed it in my final review.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const handleCloseNavMenu = () => {
+    setAnchorElNav(null);
+  };
+
+  return (
+    <AppBar position="static">
+      <Container maxWidth="xl">
+        <Toolbar disableGutters>
+          <Typography variant="h6" noWrap component="div" sx={TitleStyle}>
```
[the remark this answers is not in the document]
johnpaps13: As i said in another comment, i am fan of css-in-js and followed the ways that material ui proposes.

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,15 @@
+import Backdrop from "@mui/material/Backdrop";
+import CircularProgress from "@mui/material/CircularProgress";
+
+export default function SimpleBackdrop({ open }: { open: boolean }) {
+  return (
+    <>
```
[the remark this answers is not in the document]
johnpaps13: No, it is not. I missed it in my final review.

src/components/Header/Header.tsx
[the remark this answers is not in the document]
ngkamperlo: Yep, css-in-js doesn't require to only create styles, there is also room for classes written as json objects

src/components/ImageList/ImageList.tsx
ngkamperlo: Sorry I wasn't clear. I mean that instead of `object` you should have added an interface which describes what a catData item is about

src/components/Header/Header.tsx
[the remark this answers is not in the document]
ngkamperlo: Yes but this is a core usage concept of Typescript, don't you agree?

src/components/ImageList/ImageList.tsx
johnpaps13: I think that i need to improve my typescript skills as my exposure in it is not great yet, but i tried. I am looking forward to improve it as your colleague if possible :)

src/components/Header/Header.tsx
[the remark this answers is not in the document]
johnpaps13: Indeed, i agree. But i was used to work like this in my previous professional projects, where typescript was not used. It is just a matter of preference i suppose.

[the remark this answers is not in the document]
ngkamperlo: Unfortunately it isn't a matter of preference because browsers treat differently classes and styles to produce the final rendering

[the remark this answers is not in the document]
johnpaps13: You mean they render faster with classes, right?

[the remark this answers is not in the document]
ngkamperlo: Nope, have a look here: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useState, useEffect } from "react";
+import { Link } from "react-router-dom";
+import styled from "@emotion/styled";
+import _ from "lodash";
```
padelis, line 4: Do we need to import everything from `lodash`?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,156 @@
...
+  useEffect(() => {
+    _.includes(favouriteCatsList, selectedCatData?.url)
+      ? setFavourite(true)
+      : setFavourite(false);
+  }, [setFavourite, favouriteCatsList, selectedCatData?.url]);
+
+  const toggleFavouriteCat = () => {
+    !favourite
+      ? setFavouriteCatsList(
+          _.concat(favouriteCatsList, [selectedCatData?.url])
```
padelis, line 78: Shouldn't this be a network request, to persist favourites on all visits?

src/pages/Breeds/Breeds.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <Content data-testid="breeds-section">
+      {breedsData ? (
+        <>
+          <Table
+            breedsData={breedsData}
+            fetchCatImages={fetchCatImages}
+            handleOpen={handleOpen}
+          />
+
+          {catImages && (
```
padelis, line 107: This renders nothing when `catImages` array is empty correct?

padelis (commented):
Hello John, and thank you very much for your time and effort for this assignment.
Please have a look at my comments.

src/components/Modal/Modal.tsx
johnpaps13: I think 
import concat from 'lodash/concat';
import pull from 'lodash/pull';
could be more appropriate in order to reduce the bundle size of the import.

src/components/Modal/Modal.tsx
johnpaps13: You are right, i hadn't seen that the api has this [way](https://api.thecatapi.com/v1/favourites) to deal with the favorites so i made a custom solution. Sorry about my carelessness.

src/pages/Breeds/Breeds.tsx
johnpaps13: You are right. Although i have a console error for handling the error of fetching, i should make a catImages ? (...) : (...) in order to deal with it better.

src/components/Modal/Modal.tsx
padelis: No worries, thank you for the prompt responses 👍

src/pages/Breeds/Breeds.tsx
johnpaps13: To be precise @padelis since we are talking about array the correct is catImages.length > 0 ? (...) : (...)

src/styling/mediaQueries.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const mqCustom = (breakpointsCustom: any[]) => {
+  return facepaint(breakpointsCustom.map(bp => `@media (min-width: ${bp}px)`));
+};
+
+const mqMaxWidthCustom = (breakpointsChat: any[]) =>
+  facepaint(breakpointsChat.map(bp => `@media (max-width: ${bp}px)`));
+
+export default mq;
+export { mqCustom, mqMaxWidthCustom };
```
thsarig, line 14: Do we need those media queries inside the application?

src/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Route, Routes, BrowserRouter } from "react-router-dom";
+import "./App.css";
+import Layout from "./components/Layout/Layout";
+import Home from "./pages/Home/Home";
+import Breeds from "./pages/Breeds/Breeds";
+import FavouriteCats from "./pages/FavouriteCats/FavouriteCats";
+import NoFoundPage from "./pages/NoFoundPage/NoFoundPage";
+
+const App: FC = () => {
+  const [favouriteCatsList, setFavouriteCatsList] = useState([]);
```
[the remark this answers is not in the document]
johnpaps13: Hello Thodori, this specific state is used widely in all 3 pages and since it was only one i didn't want to use state management. Although, as i mentioned in the readme file, in the end of the day, i think that a react context approach for this would be more delicate, but i didn't have the time :( An another solution that just came into my mind, would be if i had one more layer (as general page) between App and Route Pages.

src/styling/mediaQueries.ts
johnpaps13: No, they are leftovers.

PR #21, opened 2022-02-18
Cat-challenge

maunobo:
My take on this cat-challenge


cat-app-mnb/src/App.js
```diff
@@ -0,0 +1,39 @@
...
+                cats={cats}
+                error={error}
+                onLoadMore={onLoadMore}
+              />
+            </React.Suspense>
+          }
+        />
+        {/* <Route path="/breeds" element={<Breeds />} />
+        <Route path="/breeds/:breedId" element={<BreedDetail />} />
+        <Route path="/favorites" element={<Favorites />} /> */}
```
ngkamperlo, line 32: Are those leftovers?

cat-app-mnb/src/components/modal/Modal.js
```diff
@@ -0,0 +1,41 @@
...
+
+const Modal = (props) => {
+  const catData = props.modData;
+
+  return (
+    <div className={classes.modalBackground}>
+      <div className={classes.modalContainer}>
+        <div>
+          <div className={classes.titleCloseBtn}>
+            <button onClick={() => props.closeModal(false)}>X</button>
```
ngkamperlo, line 11: Shouldn't event handler be on the top of the functional component?

cat-app-mnb/src/components/modal/Modal.module.css
```diff
@@ -0,0 +1,70 @@
+.modalBackground {
+  width: 100vw;
```
ngkamperlo, line 2: Why do you user `vw` here?

cat-app-mnb/src/lib/api.js
```diff
@@ -0,0 +1,5 @@
+export const getAllCats = (page = 1) => {
+    return fetch(
```
ngkamperlo, line 2: Are you aware of alternative libraries for http communication?

ngkamperlo (commented):
Thanks for the effort you put into this assignment mate! can you please also respond to the following questions:
1. Are you aware of typescript? have you used it before?
2. Are you aware of some UI frameworks? How come you did not use any of those?
Again thank you very much for the effort and I am looking forward to your answers!

cat-app-mnb/src/App.js
maunobo: These are indeed leftovers, routes for the other views of the application.

cat-app-mnb/src/components/modal/Modal.module.css
maunobo: This could be written as a `%` instead

cat-app-mnb/src/lib/api.js
maunobo: I am, I have used **axios** in another project

cat-app-mnb/src/components/modal/Modal.js
maunobo: This should indeed be handled on the top of the functional component, maybe as :

```
const closeModalHandler = (modalState) => {
    props.closeModal(modalState)
  }
```
Then call this function instead of directly accesing the props

maunobo:
> 

Thank you for taking the time to review my PR and for the comments, those are good catches 👍
This challenge helped me get my head around a more 'functional' way of developing, in the hope of using only plug-n-play components in it. The destination might not be on point with the missing features of breeds and favorites, but for sure it was a fun trip.

Below answers to your questions:

1. I have used TypeScript before on Angular projects and a React project. I did not use TypeScript on this project as I think it would have taken more time, which I unfortunatelly did not have. 

2. I am aware of Bootstrap, which I should have used, instead i chose to do it the hard-way.. 🤦
Styling some of the components took me a lot of time, which could be saved if I used Bootstrap, e.g the Modal.

Thanks again!

PR #22, opened 2022-02-27
Cats lover application implementation

gpositive:
Hi guys! :)

Sending you this PR that contains the basic implementation of the application described in the README file.
I'll be happy to discuss/answer any questions about it! Waiting for your feedback!

Thanks!


src/components/breedDetails/BreedDetailsContent.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { Link, useLocation } from "react-router-dom";
+import { useStyles } from "./BreedDetailsContent.style";
+
+const BreedDetailsContent = ({ data = [] }) => {
+  const classes = useStyles();
+  const location = useLocation();
+  return (
+    <div className={classes.breedDetailsWrapper}>
+      <h2 className={classes.breedDetailsHeading}>
+        {data[0]?.breeds[0]?.name || ""}
```
ngkamperlo, line 12: Could `data[0].breeds[0]` be refactored? In addition is there a possibility that data is an empty array? If yes what will the component show?

src/components/breeds/BreedsList.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+    case AsyncStatus.RESOLVED:
+      return <BreedItems items={fetchedBreeds} />;
+
+    default:
+      throw new Error("This shouldn't ever happen");
+  }
+};
+
+BreedsList.propTypes = {};
```
ngkamperlo, line 32: Is this needed?

src/components/cards/CardsWrapper.style.js
```diff
@@ -0,0 +1,9 @@
+import { createUseStyles } from "react-jss";
+
+export const useStyles = createUseStyles({
+  gridContainer: {
+    display: "grid",
+    gridGap: 10,
+    gridTemplateColumns: "repeat(auto-fit, minmax(300px, 1fr))",
```
ngkamperlo, line 7: Can you briefly explain what `fr` does?

src/components/catDetails/FavouriteToggler.style.js
```diff
@@ -0,0 +1,27 @@
+import { createUseStyles } from "react-jss";
+
+export const useStyles = createUseStyles({
+  favouriteToggler: {
+    marginTop: 15,
+  },
+  favouriteButton: {
+    backgroundColor: "#8b0000",
+    color: "#fff",
```
ngkamperlo, line 9: Maybe you could add `#fff` as a const variable somewhere?

ngkamperlo (commented):
Thanks for the time you put in this assignment mate. Can you please answer the following apart from the questions below?
1. Have you heard of typescript?
2. Are you familiar with UI frameworks? How come you didn't use any in this assignment?
Again thanks for the effort you put in this assignment. We are looking forward to your answers!

gpositive:
Thanks for the review and the questions! :) They help a lot!

I have a very basic knowledge on what Typescript is and how it helps, but I haven't ever used it tbh. I'd really like to learn more about it in the near future (it's included in the first 3 items in my 2Learn&Use list)

If by 'UI frameworks' you mean frameworks that help with html/styling architecture and provide some basic reusable components, yes. I've used earlier versions of bootstrap on production websites and Material UI in a couple of personal dev projects. 
Thought of using one of the above at first but after comparing pros + cons it turned out that it could be a bit of an overhead in terms of time, css/js bundle size and possibly styling overrides. Required styling wasn't too much at the end of the day. Even if I decided to use one of the above later it wouldn't be difficult to integrate in the current app codebase.

src/components/breeds/BreedsList.jsx
gpositive: It's a leftover. Should have removed it. Damn!
No. No props exist for this component.

src/components/cards/CardsWrapper.style.js
gpositive: > 

'fr' is just a relative unit that defines a fraction of space. Examples:
- grid-template-columns: repeat(5, 1fr); // <== Each fr takes 20% of space. (All 5 columns take 100%)
- grid-template-columns: 1fr 1fr 40% 2fr; // <== All frs take 100%-40% = 60% of space so each fr takes 15%
It can be mixed with other values as well like: %,px etc and is applied on the container width

src/components/catDetails/FavouriteToggler.style.js
gpositive: Yes. As mentioned in the README file one thought of improvement would be to use a generic color palette object.

It's not a huge app but maybe it would make sense to use a mini design system/style guide/theme config (at least to have a structure of the different colors/box shadows/borders/font rules in a central place). So a theme.config.js that defined an object much simpler than this: https://mui.com/customization/palette/ would be good I guess for starters.

Then this object would be imported in every *.style.js file that defined the JSS rules and the theming/color palette changes would be much easier.

PR #23, opened 2022-03-18
react-assignment

tbarous:


server.js
```diff
@@ -0,0 +1,26 @@
...
+    res.render('pages/app', {env});
+});
+
+app.get('/favorites', function (req, res) {
+    res.render('pages/app', {env});
+});
+
+app.listen(port, () => {
+    console.log(`Cats app listening on port ${port}`)
+})
```
ngkamperlo, line 26: Why did you create your own server?

src/components/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+const Close = styled(Times)`
+  position: absolute;
+  right: .5rem;
+  top: .5rem;
+  cursor: pointer;
+  color: white;
+  font-weight: bold;
+  width: 30px;
+  height: 30px;
+  z-index: 99999999999;
```
ngkamperlo, line 47: Why such long number?

src/components/Navbar.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const L = styled(Link)`
+  margin-right: 2rem;
+`;
+
+const Navbar = () => {
+    return (
+        <Nav>
+            <Container>
+                <div style={{display: "flex"}}>
```
ngkamperlo, line 30: Why did you use `style` at this point? could you have used something else instead?

src/pages/Breeds.tsx
```diff
@@ -0,0 +1,134 @@
...
+const CatImage = styled(ImageComponent)`
+  height: 200px;
+  margin-bottom: 1rem;
+`;
+
+const Title = styled(Text)`
+  padding: 1rem;
+`;
+
+interface Props {}
```
ngkamperlo, line 32: Is this a leftover or is it needed somewhere?

ngkamperlo (commented):
Thank you very much for the effort you put in this assignment :) Apart from the questions below, can you also please answer the following:
1. How come you used typescript instead of plan javascript?
2. Are you aware of other UI libraries ? can you list some of those?
Thanks in advance and again thank you for your time! Looking forward to your responses :)

server.js
tbarous: It gave me a better control on what pages are rendered on which routes on inital renders. Also, i had more control on the app environment by using a template engine like ejs.

src/components/Modal.tsx
tbarous: Yes this is bad practice. Normally, z-indexes would exist inside the theme for each element or seperately in an object. That way the z-index hierarchy would be more clear. Due to the nature of the project being an assignment some things are not developed in the best possible way.

src/components/Navbar.tsx
tbarous: :) Nice eye. This is the only style attribute i have used in the app. Yes, styled components could help here.

src/pages/Breeds.tsx
tbarous: For future use, probably should have been omitted.

tbarous:
> Thank you very much for the effort you put in this assignment :) Apart from the questions below, can you also please answer the following:
> 
> 1. How come you used typescript instead of plan javascript?
> 2. Are you aware of other UI libraries ? can you list some of those?
>    Thanks in advance and again thank you for your time! Looking forward to your responses :)

Thank you.

Typescript has more features than javascript and is also typed. The UI libraries i have used are:

1. bootstrap, materialize [sass]
2. vuetify [vue]
3. react-bootstrap, material-ui [react]

i used react bootstrap here only for the grid system.

PR #26, opened 2022-06-27
Feature/tsakiris challenge solution

devtsaki:
Assignment from George Tsakirs
This SPA web app has been bootstrapped with Create React App.
You can use  `yarn start` or `npm start` to run the app.
For running tests use `yarn test` or `npm test`


src/common/ducks/operations/api.ts
```diff
@@ -0,0 +1,49 @@
...
+      if (error.response?.status === 500) {
+        yield call(
+          toast.error,
+          error?.response?.data?.message ?? 'Something went wrong!'
+        );
+      }
+    } finally {
+      if (withLoader) {
+        yield put(CommonActionCreators.uiLoadingFinished());
+      }
```
ngkamperlo, line 44: Shouldn't `finally` cover all cases, i.e an `else` statement should be present too.

src/features/breeds/screens/Breeds/BreedDetailsModal/BreedDetailsModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      setCatsListPager(Constants.PAGINATION.PAGE + 1);
+    }
+  }, [getCatsByBreedName, breedName]);
+
+  const handleRefreshCatsClick = React.useCallback(() => {
+    getCatsByBreedName({
+      breedName,
+      page: catsListPager,
+      limit: Constants.PAGINATION.LIMIT,
+    });
```
ngkamperlo, line 49: What's the purpose of `useCallback` here?

src/common/components/Dialog/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './Dialog';
```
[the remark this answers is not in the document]
devtsaki: The main advantage of using a "single exit point" for a folder is that it manages the exports with a controlled way. We can export Components and sub-components from a higher level, so that we avoid complex import paths wherever they are used.
Moreover it makes us think in terms of packages, where we put together the main file, its test, any sub components, utils, constants, types and anything related with the root component, and organizing the exports in an elegant way if needed.

src/common/components/MoreButton/MoreButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  :active {
+    cursor: pointer;
+    background-color: ${Colors.primaryActive};
+  }
+
+  :disabled {
+    cursor: not-allowed;
+    background-color: ${Colors.border};
+  }
+`;
```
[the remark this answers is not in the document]
devtsaki: Generally, whenever there are many styled components, it would be better to create a `styled.ts` file and and put any styled components there. Then import them back and use them. In this case, `MoreButton` is a small file so we can just use these 2-3 styled components in the same file.

src/common/ducks/operations/api.ts
devtsaki: I'm using the condition because we want to avoid a redundant action dispatch for `yield put(CommonActionCreators.uiLoadingFinished());` if withLoader was false.
The alternative would be to yield the uiLoading on both try and error blocks, and ditch the finally.

src/features/breeds/screens/Breeds/BreedDetailsModal/BreedDetailsModal.tsx
devtsaki: Since this handler has many dependencies (catsListPager, breedName, getCatsByBreedName), it would be sane to use useCallback hook and make the handler more optimized. That means it will remain memoized if the dependencies don't change. Then the children that consume this handler will avoid redundant rerenders.

devtsaki:
1. Css frameworks have come a lot of way the past decade. Pure css files, scss, bootstrap, bulma, css modules, css in js, styled components. Nothing beats a good organized css file and carefully using only what is needed, but this is almost impossible, when having a large group of developers and complex web apps. Tailwind is an excellent framework that gives out of the box mobile first styling and excellent dead code elimination functionality.
Also in my opinion, using a headless ui kit together with tailwind, allows us to use our preferred branding style on reusable and fully third party maintenable components.
So to answer you question, I have used a lot of css frameworks and ui kits (even created one), and tailwind together with styled components is an excellent choice nowadays, at least until the next big thing comes.

2. I have been using typescript consistenly for the past 4 years. It makes the developer experience better and safer. Moreover, I remember, before typescript, we used to write tests for the prop types of a component. Using typescript correctly makes the need fot this type of unit testing obsolete and we focus more on substancial tests.

jhatzics:
Hi George! Thank you for the effort you put into this assignment. It is a really great result 💪. I would like you to give an answer to the following questions.

* Regarding security, what would you consider missing from your deliverable, to consider it production-ready?
* In case your app had to scale from several screens to hundreds, how would you approach it from an architectural perspective?

Thank you once again! You did a great job!

src/state/rootSagas.ts
```diff
@@ -0,0 +1,9 @@
+import { all } from 'redux-saga/effects';
+
+import { HomeCatsSagas } from 'features/home/ducks';
+import { BreedsSagas } from 'features/breeds/ducks';
+import { FavoritesSagas } from 'features/favorites/ducks';
+
+export default function* rootSagas() {
```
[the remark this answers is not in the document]
devtsaki: Simple Redux or Mobx, if we had to use a state management library. To be honest I could have gone with just React context and use react query for caching data responses and useReducer hooks for more "redux-like" reducers.
Last but not least, for complex applications we could utilise IndexedDB api and have an "SQL-like" data persistence.

devtsaki:
Hello Gianni!

* A must have feature is extracting environment variables (API keys or passwords) to the CI/CD configuration and not holding it inside the code. Also for production builds we could deactivate the source mapping and provide some level of obfuscation.

* About scalability. For sure we could use React lazy loading for the three basic routes. Also we could have used custom minification plugin, especially if we had a webpack configuration. The static assets of the application could have been uploaded on a cloud storage (s3 bucket or google cloud storage), so that they are available from a cdn. On really big projects we could have implemented a microfrontend architecture (this is why I prefer feature based code strructure), so that we can host each micro-frontend on a different server. There are also some server tools like varnish cache and load balancers, that optimize response times. Lastly, we could containerize the application (docker, poodman), and take advantage of kubernetes powerful features. To wrap it up, there are many tools that help us scale our application, but the most important step of scalability is scalable code. The tools come after this.

PR #27, opened 2022-07-22
Sr Implementation - Cat Lovers App

Ov3r1d3:
Future Potential Improvements -> 
 Full Suite Testing using Jest 
 Redux Toolkit Query , for the Breed Data
Full normalization of the structure


src/app/pages/Breeds/Loadable.tsx
```diff
@@ -0,0 +1,10 @@
+/**
+ * Asynchronously loads the component for Breeds List
+ */
+
+import { lazyLoad } from 'utils/loadable';
+
+export const BreedsPage = lazyLoad(
```
akontovasilakos, line 7: I like the `Loadable` component idea, but I see that being repeated in all the different pages. 
Is there any way that you can think of avoiding this repetition?
Thanks!

src/app/components/AccordionItem/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  active: boolean;
+  onClick: () => void;
+}
+
+const AccordionItem: React.FC<AccordionItemProps> = ({
+  children,
+  title,
+  onClick,
+}) => {
+  let [isActive, setIsActive] = useState<boolean>(false);
```
ngkamperlo, line 15: Why not `const` ?

src/app/pages/Breeds/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <title>Breeds</title>
+        <meta
+          name="description"
+          content="All the information for the cat breeds you want to know!"
+        />
+      </Helmet>
+      <LoadingOverlay loading={loading} />
+      <BreedModal isOpen={isModalOpen} setIsOpen={setIsModalOpen} />
+      <DefaultLayout>
+        {breeds.length > 0 ? (
```
akontovasilakos, line 52: The ternary operator is definitely a valid way of conditionally render elements in a component. Overusing it though, might affect significantly the the code readability (Disclaimer: not the case here!).
Is there any other way to conditionally render elements and have a cleaner code?

src/utils/loadable.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { lazy, Suspense } from 'react';
+
+interface Opts {
+  fallback: React.ReactNode;
+}
+type Unpromisify<T> = T extends Promise<infer P> ? P : never;
```
ngkamperlo, line 6: Why do you need this type?

src/app/pages/HomePage/CatModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+                    <div className="relative w-11/12">
+                      <ShareLinkInput
+                        type="text"
+                        id="input-group-1"
+                        value={makeShareLink(catImage.id)}
+                        readOnly
+                      />
+                    </div>
+                    <div
+                      onClick={() => {
```
ngkamperlo, line 79: Why not declare a function on the functional component level?

src/app/pages/HomePage/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  selectCatsLoading,
+} from 'store/cats/selectors';
+import { CatApiErrorType, CatApiSuccessType } from 'store/cats/types';
+import tw from 'tailwind-styled-components';
+import OptionsButton from './OptionsButton';
+import { useTranslation } from 'react-i18next';
+import { messages } from '../../messages';
+import useToastNotification from 'hooks/useToastNotification';
+
+function useQuery() {
```
ngkamperlo, line 22: Why do you have both `function` declarations and `const f = ( ) =>` declarations? Does the `const f = () =>` have a specific name?

src/app/pages/HomePage/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+          </div>
+        ) : error ? (
+          <ErrorText>{t(repoErrorText(error)())}</ErrorText>
+        ) : null}
+      </DefaultLayout>
+    </>
+  );
+}
+
+export const repoErrorText = (error: CatApiErrorType) => {
```
akontovasilakos, line 119: Could you please explain why did you decide to write this utility function here? 
I'm afraid that this decision has ended up in a circular dependency with the `useToastNotification` hook (the two files load each other).
Is this something that could be avoided?

ngkamperlo (commented):
Thank you very much for the effort you put in this assignment mate! I particularly liked the addition of multiple languages! In addition to the questions below can you please answer the following:
1. Have you used typescript in a previous role? If yes how did you like it?
2. Are you familiar with redux alternatives? have you used them professionally? 
Again thanks for the effort you put in this assignment - looking forward to your responses!

src/store/breeds/saga.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { catSliceActions as actions } from '.';
+import { Breed } from './types';
+
+/**
+ * Cat Api Request/Response handler
+ */
+export function* getBreeds() {
+  if (!CAT_API_KEY) {
+    yield put(actions.breedsApiError(CatApiErrorType.API_TOKEN_EMPTY));
```
akontovasilakos, line 14: Could you please elaborate a bit on the decision of using `yield` for your asynchronous requests?
And a follow up question here: If we were to refactor this to use async-await, how would you approach that?

akontovasilakos (commented):
Hey! Thanks a lot for taking the time and effort to submit this assignment. I hope you enjoyed working on it.

I've added some comments on your PR and we would love it if you could answer them!

Along with those comments, I have a few general question:
- Could you please elaborate on what is the purpose of the `internals` directory and how it is used within the application? I'm afraid I missed that and it.
- Why did you decide in some cases to have the icons as markup within the components? For example the flag icons where extracted as svg files. Why those were handled differently?
- If you were to test your application, how would you approach testing?

Again, thanks a lot for your effort. We really appreciate that.

Ov3r1d3:
> Thank you very much for the effort you put in this assignment mate! I particularly liked the addition of multiple languages! In addition to the questions below can you please answer the following:
> 
> 1. Have you used typescript in a previous role? If yes how did you like it?
> 2. Are you familiar with redux alternatives? have you used them professionally?
>    Again thanks for the effort you put in this assignment - looking forward to your responses!

1. I have actually been using typescript professionally for the past 1-1.5years in the react frontend that we are developing, and the truth is that I am really enjoying it. I dare say that for large projects , it is a necessity.
I really like the robustness it gives to the javascript code , since javascript can be really messy some times , plus it is very easy to make an error. 
The type inference is what actually makes typescript the way to go , because it "converts" the js language into a strongly typed language , which also makes the developer experience way better ( with autocompletes , identifying function arguments in the IDe).
A very important benefit also comes to the developer when using any npm custom packages or libraries. If the library is written correctly in typescript , with exported types , you know exactly what to expect from it. 
The only downside I see , is that sometimes I suffer from waiting the typescript compilation :(. Maybe I am using it in very large projects , with a not perfect configured webpack , but sometimes the downtime seems very high for me ( e.g. 5-8 seconds waiting for one line of code , and the ide errors not being resolved.). Still I would even agree that the cost is worth it. 

2. I would not consider them redux alternatives , but I have used professionally and developed a lot with the Apollo client and the Vuex Store. We have been actively maintaining some Vue JS projects , that are using both. The GraphQl is very powerful, once you get the hang of it after some reading , because it takes time to graps the full potential of GraphQL. In react I haven't used any other library for store management , the only thing close that we have been using is react-query , but this is not a state management tool , but a caching layer, so we have been using this in conjuction with the redux store.

src/app/components/AccordionItem/index.tsx
Ov3r1d3: True , this is wrong it should always be const, since on every rerender it is basically a new variable, so there is no reassignment on the variables .

src/utils/loadable.tsx
Ov3r1d3: The truth is that is not a code block I have written. Before starting this assignment , I searched for good boilerplates with ts , and I came across this cra template , that seemed really good for me. I actually really like the loadable part that uses the Suspense api to lazy load components , and I believe this is the way to go especially in react 18. 
Correct me if I am wrong , but this type there is only needed to infer the return type of the component that you are actually trying to lazy load , so that the function can be called correctly on the Loadable.tsx file . It does not play any special role , but only helps in extracting the return type of the promise.

src/app/pages/HomePage/CatModal/index.tsx
Ov3r1d3: It could be done there true and I think that in this case it should  indeed be done on the functional component level , you are right. It is just that sometimes when the code is one line long , we tend to insert it inside the onClick prop. But here yes a function should be extracted , since it would be more easily identifiable for anyone who reads the code

src/app/pages/HomePage/index.tsx
Ov3r1d3: When you mention specific name , do you mean the ES6 arrow function declaration ? There is no specific reason behind the different declarations in my code .To be honest , consistency was a part that I wanted to improve on my code given and this involved my function declarations. It was probably due to speed that I declared those functions with function declaration , because I used to write them in the past like that but lately I switched to arrow syntax mainly for readability , because I prefer reading this syntax.

Ov3r1d3:
> 

 

> Hey! Thanks a lot for taking the time and effort to submit this assignment. I hope you enjoyed working on it.
> 
> I've added some comments on your PR and we would love it if you could answer them!
> 
> Along with those comments, I have a few general question:
> 
> * Could you please elaborate on what is the purpose of the `internals` directory and how it is used within the application? I'm afraid I missed that and it.
> * Why did you decide in some cases to have the icons as markup within the components? For example the flag icons where extracted as svg files. Why those were handled differently?
> * If you were to test your application, how would you approach testing?
> 
> Again, thanks a lot for your effort. We really appreciate that.

1. There is actually no real reason behind the internals folder. It is not used at all in this proj. As mentioned above this whole repo was taken from a boilerplate that I used . There are some helpful utils that I intentionally left inside and did not remove them , so that I keep the boilerplate intact in case you would want to check out my source. 
2. Also , as mentioned above to ngkamperlo, consistency was the part that I wanted to improve on my code handed . At first, I imported the svgs files and this should be the way to go for every other svg file. The reason I handled the others differently , was mainly because I copied the code from the Hero Icons Page , where I took the icons I needed , as a markup. 
3. I am not as proficient as I would like in testing react applications ,since mosty we were unit testing php applications,  but I think I would mainly focus on the parts where my application would be most likely to fail. For example, the Random Cat API , (definitely create mocks for those APIs , along with error cases ) and assert that the results are being handled correctly from the app and the correct snapshot of the app is given when those apis fail. I would use the jest coverage tool to identify parts of my code that have not been tested yet and start from the most important ones , depending always on the time I have for testing the application. I do not think a full end to end testing solution would be the right case for this project since it is a small project , but maybe if the app grew bigger in the future :) .

src/app/pages/Breeds/index.tsx
Ov3r1d3: Totally , I also think that one ternary operator is fine as far as readability is 
  concerned , but nesting them could easily make the code hard to read. 
  There are a couple of ways I could think to avoid the ternary operation. One would be a component that would simply render its children , in case breeds length is > 0 , otherwise show a text . Another easy way would be to have a couple of if statements rendering the data that was needed according to the case. e.g. 
  if (breeds.length > 0)  {
    return (<Breeds />)
  }
  return (<div>Cats have no breed</div>

src/app/pages/HomePage/index.tsx
Ov3r1d3: I think that this was a decision of the moment , that indeed should be avoided .The hook should definitely not be created. The main reason I used that hook was because I wanted the toasts to be translated and since I used the react hot toast for simplicity, I created this so that I can catch the error change and grab the translation from the context. 
If this was to be redesigned , firstly I think I would create a custom toast that I would control probably from the store directly from the responses and maybe update the language configuration of the user to the store as well. 
The repo Error Text div though should definitely be a smaller component of its own , without and the function (repoErrorText) extracted to its own util ts file , since it should not be part of the component where I put it in the first place.

src/store/breeds/saga.ts
Ov3r1d3: The case was that I studied the boilerplate code and I saw that it was using redux-sagas . Redux sagas are not something that I am familiar with and I am not using them currently in the projects I am working with . I believe that these exercises are a nice way to cultivate your talent and I really enjoyed reading up on redux-sagas and how to handle the effects. Also generator functions , although I had read about them I never did actually use them in production ready code , due to their nature. I think it was a good exercise for me and tht why I picked going with the generator functions and the yield functionality. I also think it should be used becauses it adds a lot of readability to your code and no unknown effects , since always the first part of the function is taken , not leading to uknown after effects. 
In any case , if I wanted to use the classic async await , I would probably with createAsyncThunk from redux toolkit , which is what we are currently using . Pretty much the bigger part of the selectors / reducers would be the same , except for the actions part where the generator functions would be replaced with asyncThunks . THis would also allow chaining actions in the frontend with dispatch(sth()).then(()=> sth else ).finally(()=> show popup for example)

src/app/pages/HomePage/CatModal/index.tsx
ngkamperlo: Are there any other benefits on declaring on functional component scope?

Ov3r1d3: Another benefit would be the reusability of the function implements. Apart from that and the readability of the code , I would say the implementation is the same.

PR #30, opened 2022-10-02
costas gwi cats application

consmavr:


src/common/custom-card.component.tsx
```diff
@@ -0,0 +1,30 @@
+
+import React from 'react';
+import Card from '@mui/material/Card';
+import { CustonBackdropComponent } from './custom-backdrop.component';
+
+const style = {
+    position: 'absolute' as 'absolute',
```
padelis, line 7: Is `as` needed here?

padelis (commented):
Hello consmavr! 👋  
Thanks for taking the time and effort to submit this assignment.
Could you take a look at my comments below?

src/services/image.storage.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface IPhoto {
+    id: string;
+    url?: string;
+    width?: number;
+    height?: number;
+}
+
+export interface ICatPhoto extends IPhoto {
+    breeds: any[];
```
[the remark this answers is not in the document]
consmavr: Yes you are right, but first i would have to add some extra fields  in the interface  (country_code, life_span etc.) which are used in `cat-details.component.tsx`

src/common/custom-card.component.tsx
consmavr: not really just `position: 'absolute',` should be ok, just copy pasted it probably from material UI and didn't think more of it

PR #31, opened 2022-10-02
Kostas Tripalitakis cat lover challenge

ktrip96:
This is the first completed version of the cat lover app. 
It is deployed on https://platform2-0-react-challenge-cat-lover-olol.vercel.app/


cat_lover/src/App.js
```diff
@@ -0,0 +1,26 @@
...
+function App() {
+  return (
+    <CatContextProvider>
+      <Header />
+      <Routes>
+        <Route path='/' element={<Home />} />
+        <Route path='/cats' element={<Cats />} />
+        <Route path='/breeds' element={<Breeds />} />
+        <Route path='/favorites' element={<Favorite />} />
+        {/* <Route path='/favoritesv2' element={<FavoriteV2 />} /> */}
```
ngkamperlo, line 19: Is this a leftover?

cat_lover/src/components/BreedOptionList.js
```diff
@@ -0,0 +1,29 @@
...
+  const navigate = useNavigate()
+  return (
+    <div className='relative w-[300px] md:w-[600px]'>
+      <select
+        onChange={(e) => {
+          if (e.target.value !== '') {
+            openModal()
+            navigate(`?id=${e.target.value}`)
+          } else navigate('')
+        }}
```
ngkamperlo, line 14: Would it be better to have the declaration of this function on top of this functional component?

cat_lover/src/components/Carousel.js
```diff
@@ -0,0 +1,128 @@
...
+    if (carousel !== null && carousel.current !== null) {
+      carousel.current.scrollLeft = carousel.current.offsetWidth * currentIndex
+    }
+  }, [currentIndex])
+
+  useEffect(() => {
+    maxScrollWidth.current = carousel.current
+      ? carousel.current.scrollWidth - carousel.current.offsetWidth
+      : 0
+  }, [])
```
ngkamperlo, line 49: With no params, when does this `useEffect` gets called?

cat_lover/src/components/CatBox.js
```diff
@@ -0,0 +1,43 @@
...
+          className={`h-48 w-full rounded object-cover shadow-lg hover:shadow-xl hover:cursor-pointer  `}
+          onClick={openModal}
+        />
+      </Link>
+      <BsHeartFill
+        size={20}
+        className={`text-red-100 transition ease-in-out  absolute top-2 right-2  hover:cursor-pointer hover:scale-125 ${
+          containsId(favoriteCats, id) && 'text-red-500 z-10'
+        }`}
+        onClick={() => handleClick()}
```
ngkamperlo, line 30: Could this have been refactored a bit?

ngkamperlo (commented):
Thanks for the time you put in this assignment mate! In addition to the questions below can you also please answer the following:
1. Have you used typescript professionally?
2. How did you choose tailwind for this assignment?
Thanks in advance for your answers and all the effort you put in this assignment! 🤘

cat_lover/src/pages/Cats.js
```diff
@@ -0,0 +1,53 @@
...
+  }, [apiUrl, setCatData])
+
+  const handleLoadMore = (url) => {
+    setIsLoading(true)
+    fetchData(url)
+      .then((catArray) => {
+        setCatData((prev) => [...prev, ...catArray])
+        setIsLoading(false)
+      })
+      .catch((error) => console.error(error))
```
teliaz, line 27: How does this handle the UI in case of error?

cat_lover/src/components/Header.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+function Header() {
+  return (
+    <div className='flex flex-wrap gap-5 justify-center items-center font-extrabold    bg-[#363636] p-5'>
```
teliaz, line 6: Can you describe a way to declutter components from utility classes? How would you approach this?

cat_lover/src/pages/Cats.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return (
+    <div className='min-h-[90vh] h-full bg-[#363636] p-10'>
```
kioannou, line 31: During loading I can click on this button - the result is that I can spam multiple requests. Can you think of a solution?

cat_lover/src/context/CatContext.js
```diff
@@ -0,0 +1,24 @@
+import React, { createContext, useState } from 'react'
+
+const CatContext = createContext()
```
kioannou, line 3: Why did you choose this solution for saving state (react context)? Can you name an alternative?

cat_lover/src/context/CatContext.js
```diff
@@ -0,0 +1,24 @@
...
+
+// We use Context because all of React components, will need some global data
+// Doing so, we avoid the prop spreading
+
+function CatContextProvider(props) {
```
kioannou, line 8: My favourites are gone after page refresh. Was there any problem with the public API?  

The API you used, supports saving favourite cats per token. Did you had any difficulty with implementing the aforementioned functionality?

cat_lover/src/context/CatContext.js
ktrip96: Hello Kostas, I chose this approach because I wanted two children of the React Tree to communicate with each other.
Another approach would be definitely passing this data with props, but I preferred Context to avoid prop drilling.

The truth though is that in this particular application, the depth of the React Tree is not that deep. 

For e.g.  I would hold the state on App.js (there is my Router) and pass with props the favorites array as long as the setFavorite method.

cat_lover/src/context/CatContext.js
ktrip96: No, no everything was working correctly.
I had two solutions in my mind in order to keep track of the user's favorite cats.
The first one was with the public API and the second one was with the context.
I was aware that during refresh the favorites are gone. 
However I chose context just because I had already used API functionality, so I just wanted to show another way of storing the data.

I can implement the second way until our interview though if you want to. 

The functionality would be:
a. Each time a user presses the heart, send a Post request to the public API in order to store the new favorite.
b. On the first render of the favorites page, do a get request and render the favorites data.

cat_lover/src/pages/Cats.js
ktrip96: Yeap, you are right. I forgot about that. :S
I could put a ref on the button like that ` <button ref={buttonRef} onClick={handleClick}>
        Click
      </button>`

and the handleClick would be sth like that
 ```
const handleClick = event => {
    buttonRef.current.disabled = true;
    fetchData()
   .then(
    // functionality code
   buttonRef.current.disabled = false
   ) };
```

File: cat_lover/src/pages/Cats.js (hunk 5 above)
Comment (ktrip96) [reply cat_lover/src/pages/Cats.js:27]: Hello Elias 
The truth is that it doesn't :P
I made a lot of shortcuts in order to finish this task in 4 days and there are some things I've forgotten.

I can think of two ways to handle this error case.

a) Fastest way: Navigate the user to the custom error page and render the appropriate message.
b) Have a state variable like this `const [isError, setIsError] = useState(false)`
When the error is true, render the appropriate message.
```
const Component = () => {
// code 
if(isError) return <> Error Screen or Error Alert </>
return(
<div> blabla </div>
)}
```

cat_lover/src/components/Header.js
ktrip96: Yeap. I normally use styled-components on my react projects, but I wanted to try tailwind-css for the first time just to get the feeling and practice it through a project.
I don't think I would pick tailwind in a big project, but if that would be the case I would use more custom tailwind classes. 
(index.css ==> @layer ultilities{ } )

So in order to answer the question, I would pick either styled-components or emotion or the old-time classic CSS modules, in order to have a more organized CSS codebase.

cat_lover/src/components/BreedOptionList.js
ktrip96: Hello Naim. Yea, you are absolutely right.
Something like that `onChange = { e => handleClick(e)}`
And inside the handleClick method the functionality...

Far more readable and easier to maintain.

The truth is that I was in rush and I skipped this step. :S

cat_lover/src/components/CatBox.js
ktrip96: yep.
First of all `onClick={handleClick}` (my bad xd)
and I could also improve the readability of the above className. It's a little bit ugly

cat_lover/src/components/Carousel.js
ktrip96: With the empty array as above  the `useEfffect` will run once at the beginning of the lifecycle of the component and never again.
With no params (like this for example:` useEffect(()=>{})` useEffect will be called after the first render and after each update. This is its default behavior.

cat_lover/src/App.js
ktrip96: It was my pleasure, Naim. I really enjoyed this challenge.  :+1: 
1. Professionally no, I haven't used Typescript yet. I have used it though in a few personal projects. 
I have a lot of experience however with strict statically typed languages and I am looking forward to discovering the power of TS.
2. The only reason I used tailwind, was for practice. I haven't used tailwind before and I just wanted to get the feeling of it. I thought it would be acceptable as long as I don't use component libraries like Material or Bootstrap. 
At my React projects I normally use styled-components and at my Next.js projects I use module.css files for styling.

> "Is this a leftover?"

I have to let you know that I had 4 days to complete this task. I thought of two ways to store Favorite images of each user. One using Context and one using the public API. FavoritesV2.js is the API one. Due to lack of time, I had to choose one and I picked the context one. I will soon implement the second way too.

Generally, there are a lot of things I have in my mind to implement in order to make this app complete 
(e.g. lazy loading, performance issues, beautiful alerts on user actions etc.), but I wanted to make a PR with the first version of the app, to let you code review as fast as possible.

PR #34, opened 2022-10-16
Cat Lover by Theodosis Karataris

theodosiskrt:


src/App.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [breeds, setBreeds] = useState<Breed[]>([]);
+  const [favourites, setFavourites] = useState<Favourite[]>([]);
+  const [selectedCat, setSelectedCat] = useState<Cat | null>(null);
+  const [selectedBreed, setSelectedBreed] = useState<Breed | null>(null);
+
+  useEffect(() => {
+    const getInitData = async () => {
+      setLoading(true);
+
+      const breeds = await getBreedsList();
```
teliaz, line 23: Can you describe another approach to get responses faster?

src/Router/Router.tsx
```diff
@@ -0,0 +1,21 @@
...
+import BreedModal from "../components/BreedModal/BreedModal";
+import BreedsPage from "../components/BreedsPage/BreedsPage";
+import CatModal from "../components/CatModal/CatModal";
+import FavouritesPage from "../components/FavouritesPage/FavouritesPage";
+import HomePage from "../components/HomePage/HomePage";
+
+const Router = () => {
+  return (
+    <Routes>
+      <Route path="/" element={<HomePage />} />
```
teliaz, line 12: How would you handle unregistered routes?

src/api/api.ts
```diff
@@ -0,0 +1,74 @@
+import { requestSettings } from "../constants/api";
```
teliaz, line 1: Can you think of a way to abstract these file exported members?

src/App.tsx
theodosiskrt: I decided to go with this approach because I did not want to fetch the data every time a page was loaded (for example the favourites page since I am handling the changes locally). Another approach would be checking if the data for a page have already been fetched and if not, then call the api for the corresponding page.

teliaz (commented):
Thank you for submitting the test mate! Can you answer the following alongside my comments? 


- How do you value code comments and when do you think that they are necessary?


Thanks in advance and again thanks for the effort you put in this assignment!

src/Router/Router.tsx
theodosiskrt: There should have been another route with the path prop set to "*" and a not found page. I noticed a bit too late that I missed this part.

src/api/api.ts
theodosiskrt: Can you please rephrase your question? I am not sure I understand.

teliaz: Sure! Could you somehow refactor these methods? (hint DRY)

theodosiskrt: At first I thought about creating a utility that would accept the url and the settings because these look pretty repetitive, every single one of those methods has the following template:
`const response = await fetch(<url>,<settings>)
 const data = await response.json()
 return data`
 I spent some time thinking if I need to do this or not and I came to the conclusion that it is not necessary, even thought it would not hurt.

